PR #20, opened 2026-04-23
ci: automate releases via release-please

matze4u:
## What

Replace the manual/dispatched release flow with [googleapis/release-please-action](https://github.com/googleapis/release-please-action) so Conventional Commits drive versioning, changelog, tagging, and the GitHub Release automatically.

## Why

Since 1.8.0 we've had two incidents where a git tag was pushed before `task release:bump` ran, leaving `package.json` behind the tag and failing Release Guard (`1.9.0`, `1.9.1`). PR #19 added a `workflow_dispatch` pipeline as a first mitigation; release-please removes the remaining manual step entirely. Contributors no longer have to remember any release incantation — every push to `main` updates a single Release PR; merging that PR is the release.

## How it works

1. Every push to `main` runs `.github/workflows/release-please.yml`.
2. release-please scans Conventional Commits since the last tag (`1.9.1`) and maintains **one** Release PR that bumps `package.json`, `.claude-plugin/marketplace.json` (via JSONPath `$.metadata.version`), and prepends `CHANGELOG.md`.
3. Merging the Release PR tags the commit (plain numeric tag, no `v` prefix — matches existing history) and creates the GitHub Release.
4. `release-guard.yml` still runs on every pushed tag, including release-please's, as an independent cross-check.

Bump size follows commit types: `feat:` → minor · `fix:` → patch · `feat!:` / `fix!:` / `BREAKING CHANGE:` → major · `chore:` / `docs:` / `test:` / `ci:` / `refactor:` → no bump on their own (ride along).

## Changes

- `release-please-config.json` — node release type, `include-v-in-tag: false`, extra-file updater for `marketplace.json`.
- `.release-please-manifest.json` — anchors `".": "1.9.1"` so the first scan starts from the current tag.
- `.github/workflows/release-please.yml` — push-on-main trigger, `contents: write` + `pull-requests: write`, uses `GITHUB_TOKEN`.
- **Deleted** `.github/workflows/release.yml` (the workflow_dispatch pipeline from #19) — superseded.
- `Taskfile.yml` — `release:bump` kept as documented emergency fallback only.
- `CONTRIBUTING.md` — new Release section describing the release-please flow.

## Follow-ups / not in this PR

- The existing `## [Unreleased]` section in `CHANGELOG.md` is untouched. release-please will prepend new release entries **above** it. On the first real release we can either let it drift or fold its contents into the Release PR manually.
- Allow GitHub Actions to create PRs must be enabled in repo settings (`Settings → Actions → General → Workflow permissions → Allow GitHub Actions to create and approve pull requests`). Without it, release-please fails to open the Release PR.
- No PAT is used; `GITHUB_TOKEN` is sufficient unless we later want other workflows to trigger on the release tag (which they don't currently need to).

## Verification

- `task ci` locally: ✅ (184 pass, 104 warn, 0 fail — same baseline as `main`)
- YAML + JSON syntax validated
- No docs in `.augment/` / `AGENTS.md` / `README.md` mention the old release workflow, so no cross-ref cleanup needed beyond `CONTRIBUTING.md`.

## First run

The merge commit of this PR is itself `ci:` and will not trigger a Release PR. The next `feat:` or `fix:` landing on `main` will open the first Release PR; review it and merge to ship `1.9.2` / `1.10.0` accordingly.

---
Pull Request opened by [Augment Code](https://www.augmentcode.com/) with guidance from the PR author


matze4u:
Follow-up commit `879253a` adds **automated npm publish** to the same workflow so the pipeline is end-to-end after this PR.

### What's new

- `package.json` gets `publishConfig`: `access: public` + `provenance: true`.
- `.github/workflows/release-please.yml` grows a `publish-npm` job that runs **only** when `release-please` actually created a release (`outputs.release_created == true`). It checks out the tagged ref, verifies `package.json.version == tag` as a belt-and-suspenders guard, then `npm publish`.
- [npm provenance](https://docs.npmjs.com/generating-provenance-statements) is enabled via `id-token: write` and the `provenance: true` publishConfig flag — each tarball on npmjs.com will show a “Provenance” badge linking back to the exact GitHub Actions run that built it.
- `CONTRIBUTING.md` documents the publish step and the required `NPM_TOKEN` secret.

### ⚠️ Required before merge

A repo secret **`NPM_TOKEN`** must be configured, otherwise the first real release will tag + GitHub-Release successfully but fail at the publish job. Recommended: [granular access token](https://docs.npmjs.com/creating-and-viewing-access-tokens#creating-granular-access-tokens) scoped to `@event4u/agent-config`, read-write, reasonable expiry (1 year), then store under `Settings → Secrets and variables → Actions → NPM_TOKEN`.

### Failure mode

If `publish-npm` fails (missing token, npm outage, version-mismatch guard), `release-please` has already tagged the release and created the GitHub Release. Recovery is a manual `npm publish` from the tagged commit. Nothing else is at risk.

PR #268, opened 2026-05-28
fix(wizard): make voice_sample optional so setup can save

matze4u:
## Summary

Fixes a hard block in the setup wizard: on the Review step in non-draft
mode you could see **"Some fields need attention before saving."** with
the Finish button stuck, and no way to save.

**Root cause:** the welcome step pre-fills name + language; selecting a
role makes `finish()` POST an identity. But `userIdentitySchema` required
`voice_sample` (`min(1)`), and the user may never have filled a writing
sample → `POST /api/v1/wizard/finish` returned 422 VALIDATION →
"needs attention" + blocked save.

**Fix:** `voice_sample` is a tone-mirroring aid, not a prerequisite for an
identity. It is now optional — empty string allowed, defaults to `''`,
hard cap still enforced when present. The wizard never blocks a save on
it. Contract (`docs/contracts/agent-user-schema.md`) updated to mark it
optional.

Regression tests added: empty and omitted `voice_sample` both validate.

## Test plan
- [ ] CI `Node Tests` (ESLint + tsc + vitest) green
- [ ] CI `Python Tests` green
- [ ] Manual: `agent-config setup` → pick a role, leave voice sample empty → Review step saves without "needs attention"

### DIFF
--- a/docs/contracts/agent-user-schema.md
+++ b/docs/contracts/agent-user-schema.md
@@ -30,7 +30,7 @@ role:                      # required — unordered list of role labels; ≥ 1 e
   - engineer
 style:
   pace: "pragmatic"        # pragmatic | thorough | rapid
-voice_sample: |            # required — one paste of the user's typical writing
+voice_sample: |            # optional — one paste of the user's typical writing
   Mach das einfach. Wenn unklar, frag im Council.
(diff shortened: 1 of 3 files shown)


matze4u:
Superseded by #269, which includes this voice_sample fix plus the AI-Council-step removal. Closing to avoid a duplicate.

PR #1000, opened 2026-07-24
feat(reach): internet-reach operator tooling — router cancelled by its own benchmark gate

matze4u:
## What this is

A governed `internet-reach` capability area — and an honest null. The router
skill this work was scoped around **was cancelled pre-authoring by its own
benchmark gate**. What ships is the verdict-independent operator tooling.

## The gate did its job

Phase 0 was a pre-registered benchmark: 12 credential-free dev-research tasks ×
2 arms (host-native web tools vs. prototype reach prescriptions), thresholds and
three verdict bands committed **before** the run, prescriptions prototyped in
gitignored scratch first so both arms were real at scoring time. Four parallel
subagents ran the arms and judged each independently against pre-declared
acceptance evidence.

**Result: native passed 12/12; the reach arm scored 0 outright wins → `band: stop`.**
Under the pre-registered rule (reach wins only where native fails; ties are
native wins) a zero-failure native arm makes a reach win arithmetically
impossible. So Phases 4–6 (router skill, gated-platform prescriptions,
shipped-skill verdict) are `[-]` cancelled, and no capability-area claim is made.

Two reach failures were prescription defects (a 400-char excerpt cap, a `jq`
projection dropping the release body). Repairing them **cannot** move the band —
native passed those tasks too, so a repaired arm scores a tie. Nothing to
salvage by re-running, and re-running after seeing the result is the post-hoc
adjustment the pre-registration exists to forbid.

## What ships

- **`reach:doctor`** — read-only: per-channel probe status, `active_backend`,
  tier, lifecycle, and the exact **pinned** fix command for the current platform
  when a backend is missing or broken. `--deep` is the explicit network opt-in,
  refuses to run in CI, writes nothing. The table output states it is not a
  routing table.
- **`reach-channels.yml` + schema** — ordered backend candidates (swapping a
  backend is a config reorder, not a code edit), the existing four-value
  lifecycle vocabulary reused rather than a second enum forked, `last_verified`
  staleness metadata, a `removal_after` deprecation path, and an install pattern
  that rejects unpinned versions, `latest`/`main`/`HEAD` refs and archive URLs.
- **`tool_probe`** — `ok`/`missing`/`broken`/`timeout`/`error` with stale-shim
  detection (resolvable shim, dead interpreter), exit-126/127 mapping,
  timeout-only single retry, per-backend error isolation, every spawn through
  `hardenedSpawnEnv()`.
- **Three CI gates**, all fail-closed: `check-reach-channels`,
  `check-reach-prescriptions` (offline supply-chain: pinning, moving-target
  sources, intake-record coverage, pin drift, plus a scoped no-pipe-to-shell grep),
  `check-reach-staleness` (>90-day verification, deprecated-without-replacement,
  overdue removal). 13 negative fixtures each exit non-zero for their own named
  rule; controls exit 0; bad input exits 3.

## Security review — one real HIGH, found and fixed before merge

An adversarial pass (`agent-security-review` + `threat-modeling`) confirmed the
strongest property of the design and then broke it in one place.

Confirmed closed: neither spawn site uses a shell (argv arrays, `stdio: ignore`);
no reach script contains a write primitive at all; **no fetched byte reaches any
payload** — the only network-derived datum is an exit code; the credential check
`stat`s and never reads, so there is no content to hash.

**Trifecta for the `--deep` path:** private-data access is present (the
operator's token via `gh api rate_limit`) and external comms is present (one real
request) — **untrusted-content ingestion is the broken leg, broken by
construction rather than by policy**: the response body is never read and the
endpoints are compile-time constants a registry cannot redirect, so no fetched
text can steer a later action.

**The HIGH:** `reach:doctor` never validated the registry at runtime while the
schema's `probe_args` **denylist leaked** — a registry entry of `probe_cmd: sh` +
`probe_args: ["-c", "<payload>"]` passed all three gates with ✅, was then
executed, and was reported healthy under the innocuous backend label. The false
assurance was the defect: the schema claimed a probe "can never install, write,
or authenticate as a side effect", and that was untrue. Fixed here — runtime
schema enforcement in `collect()` (refuses, exit 2), a flag-shaped **allowlist**
for `probe_args`, `probe_cmd` bound to equal its backend `id`, a guarded
prototype-chain lookup, the CI refusal moved out of the CLI layer, parse errors
no longer echoing file content, and two permanent regression fixtures. The schema
comment now states a bounded claim and explicitly declines to promise upstream
binaries are side-effect-free under every flag.

## Honesty notes a reviewer should not have to dig for

- **The null is narrow.** The credential-free constraint (for reproducibility)
  excluded the two cases where a reach advantage was hypothesized: video
  subtitles (`yt-dlp` absent → `untested`, never scored as a win) and
  authenticated/rate-limited access. It says nothing about gated platforms.
- **A measured cost win is deliberately left on the table.** On all 8 tasks both
  arms solved, the reach arm cost **0.46×** the native tokens (0.26× on
  repository metadata, 0.31× on discussion search) — the native arm pays
  *discovery overhead* to find the machine-readable endpoint a prescription
  already knows. Logged outside the decision: the token threshold was authored as
  a ≤1.5× guardrail, never a win condition. Acting on it needs its own
  cost-primary pre-registration.
- **The read-only witness says only what it measures.** Its first shape
  snapshotted the whole worktree and failed reproducibly under the parallel suite
  on `agents/runtime/council/events.log` — written by a neighbouring council test,
  not by the doctor. Watching shared mutable state from a witness is the defect;
  the scope-out is documented in-file with that measurement, and a structural
  assertion (no write primitive in any reach script, with a needle check) is what
  keeps the claim honest.
- **No `CREDITS.md` entry**, deliberately: `source-confidentiality` forbids
  derivation-attribution in tracked artefacts and nothing is vendored. Provenance
  is anonymized in the archived roadmap with the link as an `ENC1:` token, and the
  reference's tokens were added to the CI denylist so future accidental
  attribution fails the build.
- **No headline claim in `docs/CLAIMS.md`** — `band: stop` means nothing is
  claimed. The two comparison rows carry the only public statements, both
  `checkable: true`.

## Verification

- `task test-ts` — **772/772 files, 8112 tests green**, twice consecutively
  (the reach set alone is 195).
- `npm run typecheck` 0 · `npx eslint <changed>` 0 · `task lint-skills` 418/418.
- `check-index`, `counts-check`, `check-condensation`, `check-md-language`,
  `check-secret-leak`, `check-refs`, `check_no_roadmap_refs`, `check_comparison`,
  `check_claims`, `roadmap:progress-check`, and all three new reach gates: exit 0.
- Hostile-registry PoC: schema exits 1, doctor exits 2, payload never executes.
- Pins verified against real registries: `yt-dlp 2026.7.4` (PyPI JSON),
  `gh 2.96.0` (local `--version`, corroborated by the release API and Homebrew).

Four e2e/UI test files failed until `npm run build` ran in this fresh worktree —
they need `dist/cli` and `dist/ui`. Measured, not assumed: after building, all
four pass. Pre-existing local reds untouched by this branch: 18
external-source hits (`CREDITS.md`, memory, archived roadmaps) and 4 untagged
roadmaps in the complexity lint.

## Roadmap

`agents/roadmaps/archive/road-to-internet-reach.md` — 47/47, archived in this PR,
20 steps `[-]` cancelled by the Phase 0 verdict with the reason inline per step.


matze4u:
## Review round 2 — resolved, with two findings that did not reproduce

The gate recommended a full council on this diff. It ran. Recording the outcome here because the recommendation is logged publicly and silence would be the wrong half of the trail.

**Council:** anthropic/claude-sonnet-4-5 + openai/gpt-4o, 2 rounds, `pr` lens, $0.13 actual. **Scope narrowed deliberately:** the full diff is ~8k lines, far past the bundle ceiling, so the council reviewed `src/scripts/reach_doctor.ts` — the one file carrying subprocess spawning, the single network mode, the credential check and the payload. That narrowing is a limitation of this review, not a claim about the rest of the diff.

### Confirmed and fixed

**`credential_path` was an arbitrary-path oracle** (both members independently called it the highest-leverage change; reproduced on the first try). A registry declaring `credential_path: ../../../../../../../etc/passwd` made the doctor report, at exit 0 inside a normal-looking health row:

```json
{"path":"/etc/passwd","present":true,"mode":"0644","group_or_world_readable":true,"checked":true}
```

Now confined to the operator's home, this repo, or the temp dir — matched against **both** the literal and the realpath'd form of every root, so a symlink cannot point out and macOS's `/var` → `/private/var` split cannot refuse the legitimate case (a bug the first attempt had; caught by running the check, not by reading it). Refused **before any syscall**, so not even existence leaks, and reported as a refusal rather than a silent skip.

**The content-echo class was open at four sites, not three.** Re-running the leak probe against *every* entry point — rather than assuming three fixes had closed the class — found a fourth: the schema validator quotes the offending value in its message, and `collect()` composed that into its refusal, so `--registry <any file>` printed that file's matching values back. All four now route through one redactor. Verified across `check_reach_channels`, `validate_reach_prescriptions` (positional and `--intake`), `check_reach_staleness` and `reach:doctor`: `AWS_SECRET_ACCESS_KEY=<redacted>`, zero occurrences of the secret, while `brew install gh`, `pipx install yt-dlp==2026.7.4` and the schema's full regex stay readable.

Also fixed: a directory handed to `credential_path` had its `0755` read as a group-readable finding (now `checked: false`, no un-actionable warning); `--deep`'s help now states what the operator consents to — a backend's rate-limit read authenticates with the credential that tool already holds, so the request presents an existing token to a third party and lands in their logs.

Both fixes are **mutation-tested**: forcing the confinement open turns six tests red (the absolute and symlink cases fail on the emitted payload, not only a unit precondition); deleting the file-type branch turns the directory case red.

### Did not reproduce — probed before believed

| Reported | Verdict |
|---|---|
| HIGH: deep-probe table not schema-gated before property access | **Not reproducible.** `collect()` validates before `load_registry`, `buildChannelRow` and every spawn; the lookup is already `Object.hasOwn`-guarded. Residual that *was* real: the exported `buildChannelRow` had no gate of its own — it now rejects malformed entries itself, and both exports name `collect()` as the gate. |
| HIGH: `sanitizeParseError` not applied uniformly across `load_registry` error paths | **Not reproducible.** A nine-way probe matrix (plain-text, valid-YAML and intake variants × four entry points) leaked nothing. `errorRow` was the one error→payload path still using a raw message; routed through the sanitizer for uniformity. |
| MEDIUM: `removal_after` compared lexicographically | **False positive twice over.** `check_reach_staleness` parses to epoch numbers via `parse_iso_date`; the schema pins both date fields to `^[0-9]{4}-[0-9]{2}-[0-9]{2}$`. The one place that *does* compare strings is sound for exactly that reason. Both facts recorded at the sites. |
| Council: newline injection via `id` into the suggestion output | **Not possible.** The schema pins channel `id` to `^[a-z][a-z0-9-]*$` and backend `id` to `^[a-z][a-z0-9_.-]*$`. |

Each refutation is now documented at the code site, so it cannot be re-raised without reading the refutation first.

### The CRITICAL from round 1

The read-only witness asserted "no write primitive in any reach script" with `source.includes(name)` — a grep. It now **parses**: `ts.createSourceFile` plus a walk, reporting a write only in callee position, with `openSync` mode-gated and a computed flag treated as unprovable (fails loud). Bite-tested both ways on the real tree: a genuine `fs.writeFileSync` fails the test naming the file and line; the same primitive in a comment plus a `'writeFileSync'` string literal passes — the false positive the grep produced.

Separately, that witness's `agents/runtime/` walk failed the full suite reproducibly on files this command cannot touch (`council/events.log`, then `mcp-telemetry/calls.jsonl` — written by other tests running in parallel). I had scoped out the first with a regex; the second proved that shape is allowlist growth. The instrument is **removed**, and the header states the narrowing rather than implying the old coverage.

**Correction to an earlier version of this comment.** It said the remaining coverage includes "the sandboxed-`cwd` runs" and a "targeted comparison" of the reach surface. Neither exists — all three `spawnSync` calls use `cwd: REPO`, and there is no reach-scoped stat comparison. I described a design I had drafted, not the code that shipped; the file's own header was accurate throughout. What the witness actually rests on: `git status --porcelain --ignored`, a `git ls-files` + (size, mtime, mode) map, an AST parse proving no write primitive exists in callee position anywhere, and the ambient-window attribution. The two risks I mis-attributed are still covered — the runs never `chdir`, so a relative-path write lands in the watched worktree; and the registry and every scanned script are tracked, so a write-back is caught on content and on mtime even byte-identical.

### Declined, with reasons

- Making `buildChannelRow` private — the export is documented as test-facing with `collect()` named as the gate; the isolation path is worth covering directly.
- A `--deep` progress indicator — output noise in a diagnostic whose per-backend deadline is already bounded and documented.

### Honesty repair on the paper trail

The roadmap's council note was titled "convergence" while recording one of four split points. All four are now recorded with how each was resolved. And both accepted costs in ADR-126 now have parked plans with explicit resume triggers instead of prose: a gated-platform bench (the credential-gated case the keyless task set structurally could not ask) and a cost-primary bench (the measured 0.46× token signal, whose pre-registration must count tokens rather than estimate them, and whose ledger must include the maintenance cost of the thing winning). Neither may reuse this roadmap's task set or thresholds.

### Verification

`task test-ts` — **773/773 files, 8167 tests**, green twice consecutively. `npm run typecheck` 0 · `eslint` 0 · all three reach gates 0 · 15 negative fixtures still exit non-zero for their own named rule · hostile-registry PoC still refused at exit 2 with no side effect.

PR #1242, opened 2026-08-10
feat: token-economy-recycling — deliberate envelope-mediated fresh starts instead of lossy auto-compaction

matze4u:
Closes `road-to-token-economy-recycling` — the third cut of the token-burn series (`dispatch` = per-spawn floor, `cache` = per-turn overhead, this = session end-of-life). A main session approaching its useful-context ceiling now ends **on purpose**: validated state flows into a recycle envelope, the user clears, the successor bootstraps from the envelope alone.

All phases landed. Phase 4.1 is recorded lapsed-optional per its blocker's own resolution clause; both blockers are resolved.

## What the measurement found (Phase 1)

First-party baseline over this machine's own store — 205 real sessions, 2026-06-25..2026-08-10. Reproduce: `./scripts-run src/scripts/session_eol_report`. Note: `agents/evidence/analysis/token-economy-recycling-phase1.md`.

| finding | measured |
|---|---|
| final context size per session | median **519,349** tokens · p90 807,937 · max 986,876 |
| auto-compact incidence | **23 of 205 sessions (11.2%)**, 31 events, **all `trigger:"auto"`, zero manual** |
| auto-compact trigger point | preTokens min **941,636** · median 1,000,551 — i.e. ~94–103% of the 1M window |
| what compaction discards | postTokens median **17,116** — the median event drops ~98% of context into one unreviewable summary, and took 123 s of paid model time in the observed case |
| late vs early turn cost | **2.1× median** (3.3× p90) billable input per call, first 20 vs last 20 calls, over 145 sessions with ≥200 calls |
| byte proxy as a fallback unit | **falsified** — Pearson r = 0.387 (n=194). Unparseable transcripts (4/205) are recorded as incidence with `null`, never a byte-derived estimate |

**Blocker `compaction-marker-shape` resolved from an observed event**, not from docs: a real auto-compaction (2026-08-06, host v2.1.222) writes `{"type":"system","subtype":"compact_boundary","compactMetadata":{trigger,preTokens,postTokens,…}}` plus a paired `isCompactSummary:true` record. Both shapes are fixture-frozen, and the scanner counts them **independently** — a host change that moves one makes the counters diverge and the report prints `marker drift: YES` instead of silently zeroing the metric.

Two claims from the roadmap header were corrected by the data rather than repeated: the trigger is not ~75% utilization on this host (it is ~94%+), and all observed compactions were 1M-window sessions — recorded as a stated threshold limitation, not smoothed over.

## The envelope (Phase 2)

`subagent_capsule.ts` gains **versioning it did not have**: the Phase-0 worker capsule shipped unversioned, so 2.1 *introduces* `CAPSULE_SCHEMA_VERSION = 2` + a `variant` discriminator additively — the unversioned worker capsule stays valid (implicit v1), and `subagent-status.json` plus the worker test suite are **byte-untouched** (5.6: one module, two variants, one validator family).

The `main_session` variant is stricter than its sibling because it has no receiver — the successor bootstraps from the envelope alone, so `task`, `acceptance_criteria`, `remaining` and the explicit `not_carried_forward` list are all mandatory. It is also the strict half of the open/closed asymmetry: the worker validator has `additionalProperties:false` at its wire schema, this variant is consumed from a *file*, so the module itself sweeps unknown keys. Consequence: **a prose summary fails twice** — as a field (`transcript_summary` → unknown key) and as content (multi-line → the shared `isShortLine`). Both fixture-proven.

- `agent-config session:recycle` — validates, writes atomically, prints the resume instruction. Registry **95 → 96** with the budget-sync triple moved in this PR, per that record's own contract.
- `handoff_context_hook` consumes it at `session_start`: validated → workspace-identity-checked (Risk 4) → 48 h staleness → injected as spotlighted DATA → **moved, not copied**.
- **Round-trip gate** (`npx vitest run recycle_roundtrip`) drives the real producer → file → consumer pipeline: session A recycles mid-task, session B resumes from the injected block alone, deliverable equals the uninterrupted control byte-for-byte. The degradation arm deletes one `decisions` entry and the deliverable **diverges** — proving the field is load-bearing, which is what "a field whose absence changes the outcome is a missing field" asks for.

## The threshold and the carrier (Phase 3 + 4.2)

`recycle-threshold-budget.json` commits **800,000 parsed tokens** — ≈ p90 of measured finals (807,937), 54% above the median healthy session, **15% below the minimum observed auto-compact trigger**. Every one of the 23 sessions that actually auto-compacted would have been advised first. Filename carries `budget`, so `lint_budget_ownership` scans it (owner + `review_by: 2026-11-10`); statically imported into the hook bundle so consumers carry the same single constant.

New `session-eol` concern on the `stop` slot: incremental counts-only per-session state, the `context-fill.json` read surface (4.2 — a future statusline finds its input waiting), and **one** advisory line **once** per session past threshold at exit-2 warn. Fail-open throughout: an unreadable transcript, a missing usage record, or `AGENT_RECYCLE_THRESHOLD_TOKENS=0` is silence, never a block.

Four registrations in `hook-token-budget.json § advisory_adoption_metrics` — the falsifiable core is pre-registered **before** data exists: if recycled sessions verify *worse* than the long-session baseline, the result is published and the threshold or the schema revises in a PR citing the numbers.

## Honest gaps, stated not back-filled

- **Post-compaction verify-fail incidence is not retroactively derivable** — transcripts carry no verify-outcome field. The forward instrument + the metric registration ship; the number does not get invented.
- **`envelope_resume_success` is audit-carried**, not deterministic — no counter exists for "had to re-ask for carried state".
- **The ≤200k-window session is not served** by an 800k threshold (it auto-compacts near ~160–206k). 11 of 201 sessions on this store; recorded in the threshold file for the review date rather than papered over with a window guess the transcript cannot support.

## Budget-file edits — additive, not a ratchet reset

Two config files carry threshold-shaped edits and the guard flagged both, so stating it explicitly: `hook-token-budget.json` gains a **new** concern row at the 1024-byte default plus four metric registrations — **no existing cap moves**; `recycle-threshold-budget.json` is a new file. `evaluator-budgets.json` 95 → 96 is the CLI-registry growth this PR causes, moved in the same PR as the record requires.

## Verification

- `npx vitest run` over the eleven touched/new suites: **87 tests green** (incl. the two `verify:` commands the roadmap declares — `checkpoint_schema`, `recycle_roundtrip`).
- `npm run typecheck` clean · `eslint` clean on all 15 changed TS files.
- `lint_hook_manifest`, `concern_registry_parity`, `concern_severity`, `check_cli_registry_budget_sync` (96 == budget == record), `lint_budget_ownership` (7 configs), `bench_hook_injection --gate` — all green.
- `./agent-config session:recycle --template` smoke-tested end to end.
- Roadmap archived via the deterministic sweep; dashboard regenerated.

Full-pipeline `task ci` is delegated to remote CI per `quality.local_auto_run: false`.

### Drive-by fix, separate commit

`fix(hooks)`: the installed pre-commit invoked `.augment/scripts/update_roadmap_progress`, which does not exist in this checkout — so every roadmap-touching commit failed with the wrong cause ("dashboard stale" when it was current). Pointed at the live `src/scripts/roadmap_progress_hook --check`, in the **source** installer, reinstalled, both branches exercised. This is the recurring stale-probe defect; fixing the installed copy alone would have left it to come back.


matze4u:
Council review, post-merge. 3 scoped slices (full diff is 3.3x the bundle ceiling), 2 members each: 5 of 6 REQUEST_CHANGES, the Stop-carrier slice split.

Must-fix, each verified in code or recomputed from the store:

1. Envelope is renamed away before its injection is delivered (`handoff_context_hook.ts:220` vs `:235`) — a crash loses it silently.
2. Two parallel `session_start` runs can both inject the same envelope.
3. `session_eol_report` never computes the cited 2.1x figure — not reproducible from the named command.
4. Undocumented `>1000` filter behind the published Pearson r: unfiltered 0.319/n=201, not 0.387/n=196.
5. "median healthy session" is undefined — median of all sessions, incl. the 23 auto-compacted ones.

Follow-up roadmap drafted; 1-2 first.

PR #1272, opened 2026-08-11
feat: enforce one authoring-contract section, reclassify the two that cannot be

matze4u:
## What this closes

`agents/roadmaps/archive/road-to-inbox-harvest-2026-08-b-authoring-contract.md` — all 15 open steps, zero deferred, archived in this PR per the roadmap PR-gate.

The subject is one defect class: **`skill-writing` labelled three sections `required` and nothing enforced any of them**, with compliance at zero. This PR turns one of the three into a real gate, reclassifies the other two honestly, and closes four adjacent gaps the roadmap bundled with it.

## Phase 1 — one gate, on a population that is a filesystem fact

`skill_linter` gains `CONDITIONAL_SKILL_SECTIONS`: a required section paired with a predicate that must be decidable from the filesystem, emitting `error/missing_conditional_section`. The first entry is `## Security constraints` when a skill directory ships `scripts/`.

The false-positive class is empty by construction, so it ships as `error` on day one rather than as an advisory that stays advisory. The four script-bearing skills carry the section **first**, so the gate lands on a compliant population — `skill_linter --all`: 437 pass, 0 fail.

Two of those sections are worth reading rather than skimming:

- **`corpus-grounding`** already had the content under `## Runtime-safety review`; it moved to the canonical heading and kept its provenance.
- **`tailwind-engineer`** records that its script **writes `tailwind.config.ts` into `cwd` on a bare invocation**, overwriting without prompting — which contradicts `skill-writing`'s own "never mutate on a bare invocation" line. Recorded with `--validate-only` named as the read-only path, plus a `## Do NOT` entry. **Changing that default is caller-visible and deliberately not in this PR.**

## Phase 1.4 — the blocker, resolved by council

`Rationalizations-to-reject` and `Non-negotiable-deliverable` have populations no property of the tree declares. AI council 2026-08-11, quorum 2/2 (anthropic, openai), both independently for **option 2: reclassify**. The decisive argument was coverage, not cost — an authored frontmatter opt-in cannot see the skill that *should* have opted in and did not, so it buys a gate without the guarantee that makes the `scripts/` predicate worth having. Both headings now read `(recommended pattern, …)`. No schema change.

What that does **not** buy is stated in the roadmap: the two patterns are now unenforced by construction and uptake will probably stay at zero. That is the honest end state for a judgement-call population; the alternative was a label claiming an enforcement nothing performed.

## Phases 2–5

- **Pattern registry** — three additions (mechanism-teaching, illustrative-not-verbatim marker, headline-metric-plus-closing-report), each about one screen. The contrastive-example slot is named in `skill-writing` **and** `rule-writing`, pointing at the six corpora that already carry those pairs.
- **Failure signatures** — a stable kebab id per row so a signature can be cited, a discrimination drill per row (several carrying the discriminator that separates neighbouring rows), and a new `stale-capability-claim` row whose first check is the live verbs, including the part that is easy to miss: `routing:doctor` prints provenance, and a `false` whose provenance is `default` records that nobody answered.
- **Claims ledger** — six external research cites bound with pointer grammar 3. **The grammar was documented with a space and the enforcing regex required the stamp to abut the URL**, so every pointer written the documented way was rejected; the form had zero usage, so nothing caught it. Fixed, both spellings pinned by tests. `superseded_by` lands **parsed and gated** (must name a real entry, rejected on a live entry, never self-referential) rather than documented-only — an unparsed field would be this ledger's own defect class.
- **Hedge-word lint** — diff-scoped per ADR-218, reusing `HEDGE_WORDS` rather than restating it, ADVISORY with the escalation trigger written before any data **and the fails-branch pre-authorised**: if the FP rate stays above 10%, the check stays advisory or is removed, it does not get a lowered bar.

## Two measurements that corrected the plan

- **Citation inventory.** The roadmap's own verify command undercounts — its pattern cannot match the `arxiv.org/abs/<id>` URL form. Corrected: **20 distinct ids across 90 lines in 32 files**, not 14 across 49 in 33. Of those, 6 are cited from `src/` where a normative sentence rests on them and are bound; the other 14 appear only in roadmap prose and stay unbound per the roadmap's own rule, recorded rather than omitted.
- **Hedge-lint FP rate.** First reading: 4 hedged lines, **4 of 4 legitimate on a human read — a 100% FP rate at n = 4**. Recorded in the roadmap rather than left for the escalation review to find.

## Completion review — nine findings, all fixed

A fresh reviewer subagent, blind to implementation context, ran against the package `dispatch_r2_reviewer` wrote (its prompt is committed at `review-input/prompt.md`). It returned **4 medium + 5 low**, committed **before** any fix, then re-bound to the post-fix scope. Four are worth the reviewer's credit:

- The hedge lint's comment predicate used a bare `*`, so **every markdown bullet counted as code** — in a gate whose only surface is `.md`. The apostrophe span alternative blanked prose between two contractions. Both suppressed the prose the gate exists to read.
- The judge-family ledger entry bound **a claim about this repo's dispatcher** to a paper that cannot back it. Narrowed, with the scope boundary written into the entry.
- The conditional predicate used `existsSync`, which a *file* named `scripts` satisfies — weaker than the `test -d` the empty-FP-class argument describes. Now `statSync().isDirectory()`.

The first of those also invalidated the measurement above, so roadmap 5.2 carries the **re-measurement at the fixed predicates** (same 4 findings, denominator 264 → 272 lines / 3,284 → 3,437 words) with the reason stated, instead of the original number standing unqualified.

## Verification

- `skill_linter --all` — 437 pass, 0 warn, 0 fail
- `check_claims` — 8 markered claims bound, ledger 60 entries (47 backed, 9 unbacked); `build-proof-check` in sync
- `vitest` — `skill_linter` 140, `check_claims` 13, `lint_hedge_words` 13
- `tsc --noEmit` clean · `task preflight` green · `check-completion-review` clean · `check_branch_freshness` current with `origin/main`

## One pre-existing red, not from this branch

`check_ci_local_parity` reports `roadmap_progress_hook` as local-only. It predates this branch (`git diff origin/main` over the hook, the workflows and `Taskfile.yml` is empty) and is left alone.


matze4u:
## Measured data for the hedge-lint escalation trigger

A parallel session ran the same roadmap and, on Phase 5, measured the lexicon before deciding whether to build the check. That measurement lands squarely on condition **(b)** of the escalation trigger declared in `lint_hedge_words.ts` — the human-read false-positive rate that has to come in at or under 10% before advisory becomes error. Posting it here rather than sitting on it, because it is the number this PR's own protocol would otherwise take 30 merged PRs to collect.

**Nothing here asks for a change to this PR.** The gate ships advisory and exits 0 unconditionally, so it cannot redden a build either way. This is input for the stage decision, not a review finding.

### Sample and method

Added `.md` lines across **150 non-merge commits** on `origin/main` (at `4fd018159`) — **10,686 added lines**. Lexicon: `HEDGE_WORDS` as exported from `bench_honesty_score.ts`, whole-word, case-insensitive. No carve-outs applied, so the figures below are an upper bound on what a carve-out-equipped run would flag.

| Lexicon | Lines flagged | Rate | Distribution |
|---|---:|---:|---|
| Full `HEDGE_WORDS` (23 words) | 79 | 0.74% | `could` 40 · `appears` 29 · `likely` 5 · `appear` 4 · `arguably` 2 · `probably` 1 |
| Epistemic-only subset (`could`/`appears`/`appear`/`likely`/`might`/`seem` removed) | 3 | 0.03% | `arguably` 2 · `probably` 1 |

### The hand read

A 40-commit sub-sample (1,359 added lines, 14 hits) was read line by line. **14 of 14 were legitimate**, and they fail in two distinct senses:

- **`could` is modal ability, not epistemic hedging** — "nobody could observe being violated", "if you could clear it with a tool", "the `stop` slot could refuse such a close". Each states a capability. Removing `could` would make them wrong, not firmer.
- **`appears` means *is present*, not *seems*** — "the count appears verbatim inside the option blocks", "a row is LIVE wherever it appears". In a tree whose prose is largely about where text occurs, this sense dominates.

Those two words are **69 of the 79 hits**, and neither is reachable by the quote-the-ban or fenced-code carve-outs: the examples above are plain prose.

### What it implies for condition (b)

On this data the measured FP rate is far above the 10% bar, so the pre-authorised branch in the header applies as written — the check stays advisory, or is removed. Worth noting the diagnosis is more specific than "the threshold was missed": the lexicon was built to score **agent output**, where `might`/`probably`/`seems` do mark epistemic hedging. It does not transfer to artefact prose about mechanisms. That is a property of the word list, not of the gate design or of the diff scoping.

The narrowed subset is the other half of the trade: it fixes precision and leaves 3 hits in 10,686 lines, one of which is a line quoting the ban itself. So a narrowed lexicon buys a defensible error rate at the cost of the signal.

### Reproduce

```bash
git log --no-merges --format=%H -n 150 origin/main |
  while read sha; do git show --format= --unified=0 --no-color "$sha"; done
# keep '+' lines under a '+++ …*.md' header, whole-word match HEDGE_WORDS
```

The full write-up — both failure senses, the per-word distribution, and a falsifier that reopens the question (a narrowed lexicon at >= 20 hits per 10,000 added lines with >= 70% precision, or one observed incident where hedged artefact prose misled a reader) — is drafted as `agents/evidence/analysis/hedge-word-lint-feasibility.md` and can land as a follow-up if it is useful to have alongside the gate.

PR #1279, opened 2026-08-11
feat(council): gate-scoped solo-attendance floor (ADR-224 implementation)

matze4u:
Implements [ADR-224](docs/decisions/ADR-224-gate-scoped-solo-attendance-floor.md)'s chosen outcome: at a **gate-class** council pass, a conclusion reached on a single present member does not carry the gate. Ordinary advisory passes keep the deliberate 1-of-2 behaviour `ceil(n/2)` chose.

Roadmap: `agents/roadmaps/road-to-council-solo-floor-implementation.md` — every step and 4 of 5 acceptance criteria closed.

## The three questions ADR-224 left open

| Question | Answer | Why not the alternative |
|---|---|---|
| What makes a pass gate-class? | **Caller-declared** — `evaluateQuorum(total, present, setting, floor)`; supplying a floor *is* the declaration | Inference silently reclassifies passes when the context changes shape, and the `pre_run` site has no `invocation` at all. A config key inherits ADR-104's user-global config — present or absent per operator machine, with no signal when absent: the exact objection that killed ADR-224 alternative (a) |
| Default for an un-instrumented call site? | **OFF** — behaviour byte-identical to before | Default-on would hold *every* pass in the repo, changing the advisory behaviour ADR-224 promised not to touch |
| Default for a declared gate-class pass? | **ON** — no second enable | Two independent opt-ins produce a mechanism overwhelmingly unlikely ever to fire |

The floor magnitude is a frozen constant, not a setting: an operator setting `min_present: 1` would disable it silently, since 1 is satisfied by every pass that met its threshold.

## Two design points that look like bugs until you read why

- **The floor clamps only its lower bound**, asymmetric to `resolveQuorumThreshold`. A *threshold* above `total` would make a pass structurally unwinnable, so it is clamped down. A *floor* above `total` being unwinnable is the floor working: a council smaller than a gate requires must hold that gate, not lower the requirement to fit the roster.
- **`heldByFloor` is a field, not a third `QuorumStatus`.** `_deserialise_quorum` hard-validates the `concluded`/`inconclusive` pair, so a third status would be breaking. Both render paths name the floor, because a hold whose own numbers met the threshold otherwise reads as an arithmetic bug.

**The floor does not branch on `isSoloConcluded`** — it compares `present` against a minimum directly, which subsumes the solo case. ADR-224's permission to branch on that predicate is therefore **unspent**, and the predicate stays advisory-only. Narrower than authorized, deliberately.

## Telemetry: schema v2 → v3

`held_by_floor` distinguishes "threshold met, floor held it" from "threshold never met" — both carry `verdict: inconclusive`, and the difference is the only evidence that could later justify or retire the floor. Every v2 field keeps its name, type and position, so a v2 reader parses a v3 line unchanged; the version moved because a reader bucketing on `verdict` alone now under-reports.

`floor_fire_rate` is registered with **no threshold committed**, and **no existing threshold is loosened** — the edit adds a metric and two honest gaps. (Stated because touching that filename trips a config-weakening guard; the guard is right to ask, and nothing was weakened.) The two rates are successor instruments: a held pass resolves `inconclusive`, so `solo` goes false and the solo rate **deflates** as the floor fires. A post-floor solo rate under 12.5 % is not evidence the degradation stopped.

## Honest scope — this ships a capability, not yet a protection

**No caller declares a gate-class pass today**, so runtime behaviour is byte-identical to `main`. An AI council (2 members, $0.05) converged 2/2 that a default-off floor with zero consumers is ceremonial compliance, and that finding is adopted rather than summarised away — its decisive premise was verified in-tree first: `legal_review_prep.require_council` appears only in the settings template, its schema, one rule and one skill, and **no code reads it**.

Two council proposals were declined with reasons, so neither returns as a new idea:

- A `gate_risk` heuristic derived from `command === 'run'` — that *is* the inference option the same review called the worst choice, and it would break `quorum.ts`'s documented purity (no context, no config).
- A CI alarm over these rates — `agents/runtime/council/events.log` is gitignored and worktree-local; two sessions read 3 and 18 `quorum_result` lines at the same moment on 2026-08-11, so a workflow would assert over nothing on a fresh checkout.

So the inertness is a **registered, falsifiable gap with a condition rather than a date**: the first caller that passes `GATE_CLASS_ATTENDANCE_FLOOR`. The honest successor is a roadmap for wiring that consumer.

## Departures from the roadmap text, recorded

**Step 2.1** asked for `min_present` to be validated in `_build_quorum`. The intent shipped, the location did not — config is the wrong carrier on a fact ADR-224 itself established (ADR-104). `QuorumSetting` is untouched and `_build_quorum` needed no change. Full list in the roadmap's Execution notes.

## Verification

- `task typecheck-ts` — rc=0 across the tree.
- `eslint` on all six changed source/test files — clean.
- `npx vitest run tests/scripts/ai_council/` — **762 passed**, 37 files.
- Full suite: 13032 passed / 34 failed across 7 files, **all pre-existing or environmental**, each identified rather than assumed. Four (`cli-e2e`, `settings.e2e`, `mcp-server.e2e`, `ui/build`) need `npm run build` for `dist/cli`; `reach_doctor` fails at worktree path depth; `tool_probe` passes in isolation (parallel-run flake); `tests/scripts/council_cli.test.ts` quota **was proven pre-existing by re-running it on a clean detached `origin/main` worktree** — it asserts the user-global council config has no quota caps, and this machine's does.
- The **full quality pipeline was intentionally not run** per a standing instruction that remote CI is the authoritative gate; the required changed-files static pass was run and is green. No claim is made about gates that did not execute.

The fifth acceptance criterion ("all quality gates pass") is deliberately left **open** until CI returns a verdict — closing it from here would drive `count_open` to zero on an unverified claim, and the roadmap archives once it is genuinely met.


matze4u:
Superseded by #1282, which shipped the same roadmap (`road-to-council-solo-floor-implementation`) and is merged; the roadmap is archived on `main`.

Timing, for the record: this PR was created 20:59:22Z, #1282 at 21:22:56Z — 23 minutes later — and #1282 merged first. Both sessions independently reached the same core finding (no gate-class consumers exist, so the floor must not be enforced today); #1282's answer is the more conservative one and additionally carries an R2 completion review, so it stands.

The one design difference is not a defect on either side and is deliberately not reopened here: #1282 ships `wouldSoloFloorHold` as a pure counterfactual, this PR enforced on caller opt-in. #1282's reasoning holds — nothing in the tree branches on `QuorumStatus`, so an enforced floor changes zero behaviour today while being able to hang an advisory pass the moment a consumer appears.

What survived from this branch is in #1283: the `min_present` provenance gap in the shipped shadow floor, plus an unrelated `pipefail`/`grep -q` bug that intermittently reds Install Script Tests. The branch stays pushed for reference.

PR #1284, opened 2026-08-12
fix(session-register): catch the duplicate-work collision at the claim, not after the PR

matze4u:
Two sessions built the same roadmap phase and opened two pull requests. Twice —
#1277/#1280 and #1280/#1281. Both times one implementation was thrown away, and
both times the coordination layer built for exactly this said nothing.

This is the ermittlung and the repair. **Five defects, each verified in the code
rather than recalled**, and the fifth is the one that would have made a fix look
finished while changing nothing.

## Why it did not fire

| # | Defect | Where |
|---|---|---|
| **D1** | The claim is scoped to the **worktree**, not the session | `session_register_hook.ts` read `<workspace_root>/agents/runtime/state/roadmap-claim.json` — one file per checkout. `sessions_cli.ts:24` said "THIS WORKTREE's session"; line 119 said "for this session". The contract contradicted itself. |
| **D2** | No staleness check on the slug | The slug was rendered raw, so a claim naming an **archived** roadmap read as an active claim. Observed: four live records carrying one identical archived slug. |
| **D3** | The collision test was `r.branch === here` — **branch only** | `session_register_hook.ts`. Two sessions on one roadmap under two branch names produced **zero** warning. This is the incident. |
| **D4** | The branch axis existed only in a memory note | `git branch`'s `+` marker caught the duplication both times, from a private note. No command carried it. |
| **D5** | The context block is emitted on `session_start` **only** | And there this session's own slug is `null` by construction, because the roadmap is picked later. **A roadmap-collision warning in that renderer could never fire on the session doing the picking.** The whole warning window sat before the decision. |

D5 is why the obvious repair — add a slug comparison to the renderer — would have
shipped a mechanism that is silent in every case it exists for.

## The repair

**The check moved to the decision.** `sessions:claim` is where a session commits to
a roadmap: foreground, output straight to the model, no hook slot, no heartbeat
delay. It now **refuses to write** a claim a live peer already holds and exits 1,
with `--force` as the deliberate escape.

That is a consistency check on its own write, **not** exclusion — the peer is not
protected and two simultaneous claims still race. The register's contract forbids
building exclusion on it and that limit is restated wherever the refusal is
described. What it does buy: the second session to arrive cannot record a claim it
cannot honour, and cannot miss the notice.

**It reads two sources, because either is silent alone.**

- **Claim axis** — a live record with the same slug. Needs `sessions:claim` to have
  run, which is model-carried.
- **Branch axis** — an unmerged branch in another worktree whose name carries the
  roadmap's distinctive **tail**. Needs nothing but a checkout on disk. The tail,
  not the whole slug: neither colliding branch in either incident contained the
  slug, and `dispatch-safety` is what a branch name actually carries.

**Live proof.** On this repository the refusal fires and names
`feat/dispatch-safety-confirmation` — the branch of the PR that was duplicated.
Had this existed, the collision would have surfaced at claim time.

Plus: the claim is keyed on the session id (legacy files still migrate); stale
slugs are labelled `← STALE` and treated as absent everywhere; the renderer reports
roadmap-before-branch because one means duplicated work and the other means two
sessions can see each other's commits; and `sessions:list` prints the branch axis,
filtered to unmerged branches in one extra git call — unfiltered it listed 30+
long-merged worktrees here, and a list that is mostly noise is a list nobody reads.

## Limits, printed rather than hidden

- A branch created seconds ago has no commits, so it counts as merged and does not
  appear on the branch axis. That first minute belongs to the claim axis.
- The name match is a heuristic — a reason to look, not a verdict.
- Two sessions picking simultaneously still race. No advisory layer fixes that, and
  the register header says so.
- `sessions:claim` remains model-carried. Nothing forces the invocation; what
  changed is that invoking it can now fail loudly instead of silently recording a
  duplicate. The command text now states the consequence: skipping the claim does
  not merely delay visibility, it removes the only axis that sees a peer's first
  minute.

## Verification

| Check | Result |
|---|---|
| `tests/scripts/session_register.test.ts` | 54 pass (was 22) |
| `tests/server/schemas/parity.test.ts`, `lint_settings_classes` | pass |
| `task typecheck-ts` | clean |
| `eslint` on the changed set | clean |
| `task preflight` | green, no regressions |
| `lint_hidden_unicode`, `lint_skill_frontmatter_safety` | clean — the two agent-security gates preflight does **not** read |
| `build:install-bundle` drift | none |
| `sessions:claim` against the real repo | exits 1, names the duplicated branch |

Every test group leads with the input that produced no warning before. Two
production defects were found **by** the tests: the env-var precedence was
backwards (`AGENT_CONFIG_SESSION_ID` lost to the ambient `CLAUDE_CODE_SESSION_ID`,
so the package's own override could not override), and the worktree self-check
needed realpath (macOS `/var` vs `/private/var` would have made every session
report itself as a collision).

## The other half of the ask — #1280

Read live, not assumed: **#1280 is `CONFLICTING`/`DIRTY`** since #1281 merged, and
it carries **no substance `main` lacks**. Its `hooks:status --pending` is step 2.3
on a different verb than the `gates --pending` that shipped, and two enumeration
surfaces for one thing is a defect rather than a gain. It converged independently
on the same `fs.renameSync` atomic claim, and has no TTL/expiry and no 4.3.
Closing it is a maintainer action, so it is named here rather than done.

**Still not done:** the R2 completion review. It needs a fresh reviewer *subagent*,
which this session is instructed not to dispatch; the unfilled skeleton is
deliberately not committed, because a committed skeleton reads as a review that ran
and filling it myself would be the author authoring the verdict.


matze4u:
Independent review of this diff at `41b2f699` — I reached the same diagnosis from the other side of the same collision (my session was the one holding the invisible `road-to-council-solo-floor-implementation` claim while four records showed `…-dispatch-safety`), built the per-session keying, then found this PR already open and 20 minutes earlier. So this is a review, not a competing change: the keying here is right and broader than mine, and `claim_is_stale` covers a case I did not have.

One residual hole, verified against this diff rather than inferred.

## A legacy claim file with no `session_id` is still returned to every session

```ts
if (legacy.session_id !== undefined && id !== '' && legacy.session_id !== id) {
    return null;
}
return legacy.slug;
```

When the legacy file carries **no** `session_id` — which is every file written by the pre-change `cmd_claim`, since it never wrote one — the first conjunct is false, the guard is skipped, and `legacy.slug` is returned for **any** caller. That is exactly the mis-attribution this PR removes, preserved for precisely the files that caused it.

The docstring states the intent as *"the legacy file counts only when it carries no `session_id`, or one that matches"*, so this is a deliberate back-compat choice rather than an oversight. The point of this comment is only that its cost is larger than it looks, for two reasons:

1. **`claim_is_stale` does not cover it.** That filter catches an archived slug. A legacy file holding a **still-active** slug passes the filter and keeps being credited to every session in the workspace.
2. **Nothing ever removes the file.** `--release` unlinks the per-session path, so the legacy file survives every release, every session end, and every upgrade. The window is not "until the next claim" — it is unbounded.

This is not hypothetical for this repository right now:

```
$ cat agents/runtime/state/roadmap-claim.json
{ "slug": "road-to-inbox-harvest-2026-08-b-dispatch-safety", ... }   # no session_id
$ git ls-tree --name-only origin/main agents/roadmaps/ | grep -c dispatch-safety
1                                                                    # still active
```

So after this merges, that file keeps putting one slug on every record in this checkout — the symptom the PR body describes — until someone deletes it by hand.

## Three ways out, cheapest first

- **Ignore a legacy file that carries no `session_id`.** One clause. A session that claimed under the old scheme loses that claim once and re-runs `sessions:claim`; an absent claim is recoverable in one command, a wrongly-credited one is not. (This is what I had chosen, so treat it as a disclosed preference rather than a neutral recommendation.)
- **Consume it on first read** — return it once, then unlink, so at most one session inherits it and the file cannot outlive the migration.
- **Keep the read, delete the file in `--release` and on `session_end`**, which bounds the lifetime without changing the read semantics.

Any of the three closes it; the current shape is the only one where the old behaviour is permanent.

## Smaller notes

- `read_claimed_slug(workspace_root, session_id?)` makes the id optional, so a caller that forgets it silently gets legacy semantics instead of a type error. A required parameter would make the one call site that matters impossible to get wrong — there is exactly one, in `build_record`.
- Worth a test either way: *a legacy file without `session_id` is not credited to a session that did not write it*. The claim path had no coverage at all before this PR, which is how the original scoping bug shipped; that specific assertion is the one that would have caught this residual case.

matze4u:
Addressed — and your residual finding was the right one to lead with. An R2 completion review over the same diff had reached it independently as its finding 6, so it arrived from two directions in the same hour.

## The legacy-file hole

Fixed with your **option 1**, and slightly wider than the clause you quoted. The condition is no longer a comparison of ids but a fact about who writes the file:

```ts
// a session that CAN identify itself never writes the legacy path,
// so a legacy file it did not write is not its claim
if (id !== '') {
    return null;
}
return legacy.slug;
```

Your reading of the old guard was exactly right, including the part I had not seen: `cmd_claim` serialises `session_id: null` on that path and `_read_claim_file` drops a null, so `legacy.session_id !== undefined` was **never** true. The guard was unreachable for every file this codebase writes — the back-compat clause was not a deliberate trade, it was dead code, and the test that "covered" it hand-wrote a record production cannot produce.

The cost you asked me to weigh is now asserted rather than hidden: `the pre-upgrade claim is DROPPED for an identified session — the price, stated`. Losing a claim is one command to restore; crediting a peer's work is not.

## Your point 2 — nothing ever removed the file

This was still open and is the more useful half. A session now unlinks whichever claim file it actually writes, at `session_end`:

- identified session → its per-session file (which also closes the unbounded-growth defect R2 raised as finding 3: one file per session id, retained forever);
- id-less session → the shared file, which it owns by construction.

An identified session still never touches the shared file — it may be a peer's, which is also why `--release` no longer unlinks it (R2 finding 10: it was silently destroying a peer's claim in a mixed-host checkout).

So the file in this repo you pasted stops being credited to anything the moment this merges, and it is removed by the next id-less session end rather than surviving indefinitely.

## Your smaller notes

- **`read_claimed_slug`'s id is now required.** You are right that optional meant a forgetful caller got legacy semantics instead of a type error. `null` stays legal — it is the host-exports-nothing case — but it has to be passed. One call site, `build_record`, as you said.
- **The test you asked for exists**: `a legacy file with NO session_id is not credited to a session`, written with the real shape (`{slug}` and nothing else, which is what the pre-change `cmd_claim` produced). Your reason for wanting it specifically is recorded in the test body — the claim path had no coverage at all before this branch, which is how the original scoping bug shipped.

## Also in, from the adversarial gate on the same diff

- **Path containment instead of caller honesty** (its two blocking findings). `roadmap_claim_rel` asserted nothing locally and leaned on `safe_stem` in another module; it now checks that the joined path stays under the state directory. `claim_is_stale` replaced its character blacklist with the question that matters — would this read leave `agents/roadmaps/` — which covers traversal, an absolute path and an unenumerated encoding with one check.
- **Token-boundary branch matching.** `dispatch-safety` is a substring of `redispatch-safety-valve`. The tail now needs a delimiter or an end on both sides.
- **`Measured twice` is backed** — both incidents cited by PR number, branch pair and merge SHA in `parallel-sessions.md`, since you and the gate both flagged it as an unbacked headline.

## One thing I did not take

Your first option was to ignore a legacy file without a `session_id`; I ignore it for any identified session, `session_id` present or not. Same direction, one clause fewer, and it removes the case where a legacy file written by a *future* id-less peer would still be inherited.

All 15 R2 findings are dispositioned `fixed` in `agents/evidence/reviews/feat-parallel-session-collision-hardening.findings.md`, re-bound to the post-fix scope. 66 specs on the register path, `check-completion-review` clean.

Not resolvable via the UI, for the record: this is a PR-level comment rather than a review thread on a line, so GitHub offers no Resolve button for it — this reply is the disposition.

PR #1340, opened 2026-08-13
feat(settings): orchestration is standing-authorized — worktrees and council move from per-run approval to a ceiling

matze4u:
## What

Orchestration is standing-authorized. The bound moves from a per-run approval to a ceiling.

| Surface | Before | After |
|---|---|---|
| `worktrees.mode` | `ask` — permission ask on every worktree creation | `on` — standing permission; ignore-check + clean-baseline gates unchanged |
| `ai-council` Procedure § 3 | unconditional "Confirm spend … require an explicit `1`" | three cases: no billable member → no gate · billable **with** a ceiling → no ask, `on_overrun` still fires on breach · billable with **no** ceiling → ask |

Recorded as [ADR-229](docs/decisions/ADR-229-orchestration-is-standing-authorized.md).

## Two things deliberately NOT changed

**`subagents.adversarial_council` stays `off`.** This was in the original request and is declined on evidence. The pre-registered claim resolved as an **honest null on 2026-07-21** (`docs/benchmark.md`): single-skeptic residual recall 0.60, 2-vendor panel 0.60 — **zero lift**, the second vendor's catches a strict subset of the first's — and a **100% false-positive rate** on all three controversial-but-correct controls. Per the locked pre-registration gate the surface stays default-off permanently. It is a feature switch over a measured-negative capability, not an approval gate; `on` would spend money to flag correct code on every high-risk change, and `ask` would be an approval burden with a negative expected payoff. Its `describe` text was corrected — it previously implied a pending benchmark.

**`ai_team.allow_delegate` stays `false`.** It grants an external tool write access to the repository. Different axis, scoped out.

## Notes for the reviewer

- **A permissive default was made permissive on purpose.** `settings-ask-protocol` carries *"A SILENT PERMISSIVE DEFAULT IS A DECISION TAKEN IN THE USER'S NAME. NEVER TAKE ONE."* That Iron Law binds the **agent** picking a default mid-run; this is a maintainer changing a shipped default, which is why the ADR exists rather than a bare diff.
- The class-B conservative-default invariant is untouched: both keys are class C, and that invariant is scoped to B (`settings-classes.md` § 88). Verified no linter enforces conservative defaults for C.
- `mode: "on"` is quoted deliberately — bare `on` parses as boolean `true` and `validate_agent_settings` rejects it. The failure was hit and fixed, and the reason is now a comment in the template.
- The `jit_ask_budget` ratchet fired exactly as its own comment predicted ("becomes a comfortable fiction after the next default flip"). The assertion is inverted rather than deleted, and now covers all four C-ask keys instead of one.
- One edit could not be applied: `src/domains/meta/council/default/command.md` § 3 still reads **"Cost confirmation — ALWAYS ASK for billable members"**. The auto-mode classifier blocked the write. That section needs the same three-case treatment as the skill; until it lands, the command file and the skill disagree.

## Verification

- `task preflight` — green (after merging `origin/main`)
- `npm run typecheck` — green
- `validate_agent_settings` — green (caught the YAML boolean, then green)
- `check_bundle_path_leakage` — green; install bundle rebuilt with a clonefile `node_modules`, diff is 3 lines, no worktree paths
- 134 tests across `jit_ask_budget`, `settings_defaults_layer`, `validate_agent_settings`, `self_repair_deny_list`, `team_dispatch` — all passing


matze4u:
Superseded, split in two on the maintainer's work order.

The worktree half decided the **opposite** of what is wanted — it flipped `worktrees.mode` `ask → on` (standing authorization). The requirement is instruction-only: the agent must never create a worktree it was not asked for. #1343 deletes the key instead of re-defaulting it, and also defuses the session-register block that was the actual source of the withheld-commit behaviour.

The council spend-ceiling half was orthogonal and survives unchanged as #1342.

ADR-229 is rewritten in #1343 to record the instruction-only doctrine and the withdrawn standing-authorization alternative; the council decision moved to ADR-230 in #1342. Nothing from this branch is lost.

PR #1437, opened 2026-08-19
feat(hooks): per-turn injection aggregate + nudge-interference fixture

matze4u:
Closes Phase 4 of `road-to-standing-context-40k` — steps 4.1 and 4.2, AC-4 met.

## 4.1 — the per-turn aggregate, with runtime teeth

`src/config/hook-token-budget.json` capped a concern's emission per fire and
the cross-concern sum per slot per fire. Two gaps, and the step's wording
understated both:

- **No ceiling on a turn.** `pre_tool_use` / `post_tool_use` fire once per tool
  call, so the per-slot rows bounded one fire of a repeating axis.
- **No runtime consumer at all.** `bench_hook_injection` was the file's only
  reader. Every cap that shipped in 3.1 was CI-only over committed fixtures;
  `dispatch_hook.ts` never imported the budget.

This adds `per_turn_aggregate_cap_bytes` plus the enforcement half in
`hooks/turn_injection_budget.ts`, wired at
`dispatch_hook._apply_turn_injection_cap`. `cap_bytes: 10240` is **derived, not
measured**, and the config says so: it is the sum of the committed per-slot caps
for a turn with exactly one tool call. No live per-turn distribution exists; the
injection census revises it at the review date.

Risk 5 of the roadmap ("the drop policy hides a safety-relevant warning") is
answered by construction: droppability is `NOT blocking AND NOT fail_closed`,
stated negatively so a severity tier added later stays droppable rather than
silently inheriting an exemption. An emission that stays over cap is emitted in
full and the overflow is recorded — an over-cap warning is a budget finding, a
swallowed block is a safety failure, and the two are not tradeable.

**Note for review — this is a ratchet ADDITION, not a threshold loosening.** The
`block-config-weakening` guard warns on any edit to this file. No existing
number moved; a previously unbounded axis gained a cap.

## 4.2 — the interference fixture, and what it found

The `user_prompt_submit` slot runs ten concerns on `claude`, two of which read
the same prompt and inject an advisory line. Nothing asserted they stay out of
each other's way.

Measured before authoring, over the widest prompt corpus in the repo — the 510
prompt lines of `tests/eval/routing-matrix`: 3 fired `delegation-nudge` only, 44
`skill-route` only, 463 neither, and **0 both**. So a fixture built only from
that corpus would have been green over a case it never exercises.

**Two hand-constructed prompts fire both, reproducibly** — enumerated files plus
strong single-skill terms in one sentence. The interference the step names as
the first pair to pin is live, not hypothetical. The corpus records it and is
bidirectional: a `max_nudges` class going above its number reds, and the
recorded `overlap` class going below its number reds too.

**Not fixed here, deliberately.** Suppressing the second nudge is a behaviour
change to two shipped concerns and needs a judgement about which one yields;
4.2 asks for the instrument. The follow-up is recorded under the phase.

## R2 completion review — 10 findings, all fixed

A fresh reviewer subagent found 2 high, 4 medium, 4 low against the first
implementation. The artefact is committed **before** the fixes, so what the
review found is auditable independently of what survived it.

The two highs were real defects in the turn boundary:

1. A host binding no `user_prompt_submit` never reset the accumulator, so the
   per-turn cap silently became a per-**session** one. `augment` binds three
   non-exempt slots and no prompt slot, and past the cap every advisory would be
   dropped for the rest of the session. `hasTurnBoundary` is now resolved from
   the manifest, and the roadmap note that wrongly claimed the module already
   failed open is corrected rather than left standing.
2. The empty-message early return skipped the reset — and by this branch's own
   corpus that is 463 of 510 prompts, i.e. the common path. The first test suite
   pinned the early return as intended, which is why it could not see the leak.

The mediums are worth naming because two are honesty repairs rather than bugs:
a crashed concern was charged for bytes `emitFor` never emits; a dropped message
still counts as a fire in `rule-trips.json`, so the row now states that `fired`
means `fired, possibly undelivered` and names the counter to subtract instead of
silently redefining five pre-registered numerators; the bench comment no longer
claims its fixture figure is the number the runtime enforces.

## Verification

- `tsc --noEmit` clean; 492 tests in `tests/scripts/hooks` green (40 of them new).
- `bench_hook_injection` reports the aggregate line and reds on breach.
- `check_completion_review` clean; all ten dispositions `fixed` with the
  repairing commit in the Ref column.
- `task preflight` green.

One deliberate omission: the AC note quotes **no** byte figure. It cited
`1140 B`, and merging `origin/main` moved it to `922 B` the same day because a
concern this branch never touched stopped firing under the fixture — the number
belongs to the fixture set, not to this phase.


matze4u:
Closing as **superseded**, not rejected — per an AI council decision (2/2 seats)
after a detailed side-by-side comparison against #1439, which merged the same
roadmap phase (`road-to-standing-context-40k` Phase 4, steps 4.1 and 4.2) about
an hour before this branch was ready.

Both seats chose this over reverting the merged implementation, and the reason is
specific rather than a coin-flip on "who was first": #1439 carries behaviour this
branch does not have at all — nudge arbitration (`nudge_rank` wired into the
manifest, an eviction order, a futility check that refuses to drop when dropping
cannot reach the cap), discriminated drop reasons, ceiling-cause attribution,
`hooks_doctor` integration, a contract-doc update, and end-to-end proof through
the real dispatcher. This branch's advantages were orthogonal *corrections*, which
are portable; the missing behaviour is not recoverable without rebuilding #1439.

Two claims from this branch also did not survive their own evidence, and both were
mine: the compiled-manifest fallback is not reachable in a supported install
(`src/scripts/hook_manifest.json` is tracked and inside the npm `files` list), and
the absent-`severity` exemption is forward-compatibility only (0 of the shipped
concerns omit the key). Its 10,240-byte ceiling was worse than #1439's: a
two-tool-call turn breaches it while breaching no per-slot row.

What was genuinely separable has landed in **#1441**:

- the injection ceiling charging bytes an `allow` verdict never emits — measured,
  not read: a 9,000-byte rc-0 message emitted 0 B and charged 9,000 B;
- one turn counter per workspace instead of per session, so two concurrent
  sessions silently disabled the cap for each other;
- the bench's "deterministic, CI-reproducible" claim, and the AC note that had
  pinned a reading which moves with the working tree.

The review history here stays worth reading: `agents/evidence/reviews/feat-standing-context-40k-phase4.findings.md`
on this branch records 10 findings with their dispositions, and two of them
(the turn-boundary gap and the empty-message reset) are the same defect class the
R2 review of #1441 caught again in a new shape — which is why that artefact is
linked rather than discarded.

PR #1479, opened 2026-08-20
roadmap: archive the three completed roadmaps that merged unarchived

matze4u:
`roadmap:progress-check` is **red on `main`** right now. Three roadmaps reached zero open steps and merged before the archival sweep ran, so the trunk carries them complete-but-unarchived:

```
❌  Completed roadmaps are still in `agents/roadmaps/`:
      - road-to-hook-state-followups.md  (10/10 done)
      - road-to-maintainer-bus-factor.md  (7/7 done)
      - road-to-release-review-p0.md  (12/12 done)
```

This is the deterministic sweep run over all three, plus the archive index, the dashboard, and one migrated inbound reference. Dashboard: **29 roadmaps, 273/549 steps** (from 32).

## One inbound reference deliberately not rewritten

The sweep migrated one reference automatically and left a second broken:

```
🔴 agents/evidence/reviews/evidence-artifact-typing.findings.md:29
   — path `agents/roadmaps/road-to-release-review-p0.md`
```

That line is a **table cell inside a scope-bound findings artefact** — `<!-- completion-review: v1 | scope: beaed2083… -->`, type `rebind-event` — recording what a round-1 finding said about the tree at that scope. Rewriting it to point at the new path would falsify what was actually reviewed, which is exactly the failure `docs/contracts/evidence-artifact-types.md` distinguishes types to prevent.

So the line carries the sanctioned per-line `<!-- ref-ignore -->` marker instead, which is the escape the reference gate documents for this case. The gate is green on a recorded exception rather than on a rewritten record. Note that the artefact's `context-manifest` block names the same path and was **not** flagged — manifests are outside the gate's walk, and it stays untouched for the same reason.

## Gates

`check_references`, `check_no_roadmap_refs`, `lint_empty_roadmaps`, `lint_roadmap_later_disposition`, `check_roadmap_trackable`, `lint_roadmap_blockers` and `roadmap:progress-check` all pass, plus the pre-push preflight. `check_references` was red before the marker and green after — verified in both directions rather than assumed.

## Why this is its own PR

Archival belongs in the PR that completes a roadmap, and for these three it did not happen — the drain run's roadmap PRs were deliberately scoped to closing the work, with archival held back so a reviewer could see the closure before the file moved. That was the wrong split for the trunk's benefit: it left a required-ish check red on `main` for anyone else's branch to inherit. Later drain PRs in this run carry the same shape, so this sweep will likely need one more pass at the end.


matze4u:
Superseded: every artefact on this branch is already on `main`, so the effective diff against `main` is empty (`git diff origin/main...HEAD` returns nothing).

- the three archive moves (road-to-hook-state-followups, road-to-maintainer-bus-factor, road-to-release-review-p0) landed via #1493, whose regen archived a superset of six roadmaps;
- `agents/evidence/reviews/evidence-artifact-typing.findings.md` landed earlier via #1391 (`097ab6549`), not via this branch.

Closed rather than merged per the drain-run rule that an empty PR is never merged to raise the count.

PR #1495, opened 2026-08-20
roadmap: complete road-to-per-turn-hook-economy

matze4u:
All six council decisions applied — four as shipped code, one as a decline, one as a decision *not* to set a number — and Phase 0 transferred to a host-repro stub. Two items stay `[~]` by decision, which holds archival under Iron Law 3 and is the mechanism working rather than a gap.

## All six option letters matched, with two notes

Verified against each blocker's own option set rather than a paraphrase: `b-guard-tool-partition` (c), `b-injection-scan-unwrap-security` (a), `b-payload-read-parse-dominates` (a), `b-per-turn-composite-bar` (b), `b-stdin-read-failure-policy` (c), `b-stop-async-split-prerequisites` (a).

Two places where the brief and the file differ, stated rather than forced:

- **`b-guard-tool-partition` option (b)** was described to me as "advisory-only"; the file's (b) is *"partition only the advisory concerns and keep one unscoped group for the guards, which keeps the guards safe and buys **nothing**"* — a partial partition, not a decline. Both seats refuse to ship it; (c) adopted.
- **`b-stdin-read-failure-policy`** — the brief's wording is the *intersection* of the file's (a) condition with (c)'s slot restriction. The file makes that explicit: "(c) is (a) narrowed to where a guard can actually refuse". Implemented as the conjunction.

## Implemented versus recorded

**Implemented:** the stdin deny (c), the injection-scan contract and unwrap fix (a), the read-exit measurement cell (a), and P3 of the async sequence.

**Recorded only:** the guard-tool partition is a decline, so there is no code by design. The composite bar is a decision plus an arming record — the row already shipped `observe_only: true`, so what landed is `observe_only_decision`, `observe_only_since_version 14.6.0`, a four-step `arming_procedure`, and an `arming_precondition` (≥ 10 CI readings / ≥ 2 sessions) marked as a **stated** minimum rather than a measured one.

**Partly pre-empted:** P4 was already closed on the trunk by `bcbb0380b` — checked against the merged tree (the old code computed `Date.now() - start > 5000`; the new code stats the companion's mtime against a 30 s bound). The roadmap is corrected in three places rather than shipping the false claim.

## Sabotage probes, and two harness defects they exposed

| Probe | Result |
|---|---|
| stdin: never deny (pre-fix) | 3 cases red |
| stdin: deny everywhere (option a) | 4 **different** cases red |
| stdin: restore `''`-on-failure | 4 cases red |
| injection-scan: restore the pre-fix unwrap | 11 of 14 red |
| dispatch-issues: unlocked append | **3 of 96 lines survived** |
| `update_json_under_lock`: read outside the lock | **3 of 24 increments, 1 of 8 rollups survived** — re-run after the merge against the *new* primitive, same result |

The first concurrency test **passed against a sabotaged lock**, because the barrier was written after spawning (children raced tsx startup, not each other) and the mutator finished in microseconds. Fixed with a wall-clock barrier deadline and a 25 ms hold inside the mutator — workload, not code under test.

## The merge hid a real divergence

`bcbb0380b` changed the RMW primitive under this branch — a **file**-keyed lock instead of directory-keyed, and a `written`/`skipped`/`failed` return instead of a boolean — and the merge was **clean**. This branch's `update_text_under_lock` still held the directory lock and still returned a boolean, and its caller used `if (!landed)`, which would have been a **silent no-op** against the union since every member is truthy. Converged in its own commit, the caller compares the literal, the contract doc is updated (main documented the file-keyed lock for the JSON path only), and P3's regression was re-verified against the merged primitive.

## The transfer

`agents/roadmaps/stubs/road-to-per-turn-hook-economy-host-repro.md` carries 0.1–0.3 verbatim (including 0.2's fixture sizes and its declared else-branch), four moved items, and a producer only one person can be: **the reporter of the original slowdown, on their own machine, at the version installed when they reported it** — nobody else can supply it, because the report is what defines "affected". Step 0.0 explicitly does **not** move: it already ran here, and its result is why the rest is a transfer rather than a cancellation.

Three readings taken here as the **wrong-machine control**, so a later reader can separate movement from noise:

| Probe | Reading, this machine |
|---|---|
| `AGENT_CONFIG_HOOKS_ISOLATED` | **unset** — 7 tracked mentions, all code and docs, no shell profile, no CI export. The flag hypothesis is **untested here, not refuted** |
| Installed version | **14.6.0** (node v26.7.0, Darwin 24.6.0, arm64) — not the report's version |
| Turn-end-gate refusal records | **0** — and the only state file is a May smoke fixture whose schema carries **no refusal counter at all**, so 0.3 needs a counter added before it can be censused |

Gate 1 can close the whole thing alone: the flag set anywhere in that environment explains the report and retires 0.2 unrun. Gate 2 is an `AND` — a matrix against a *guessed* prior version produces a number nobody can attribute.

The § 3 finding travels **into** the stub: the regression does not reproduce at the dispatcher level, the likelier cause is the mixed-trigger activation flip, and the **refuted** `grep -l '^paths:'` discriminator goes with it — a promoter reaching for that one-liner gets a confident wrong answer.

## Estate accounting

`open_blockers` 64 → **58**, re-anchored at the merge: main had walked to 64 for its own closures while this branch walked to 59, and 58 is what the merged tree reads. Every history entry from both sides preserved (34 entries), none collapsed.

The −6 is **not six pieces of finished work**: four blockers discharged by code, one a decline, one a decision not to set a number. The transfer itself moves **no** metric, because `stubs/` is counted by neither `active_roadmaps` nor `open_blockers`.

Recorded rather than deduplicated: the trunk reached the **same six decisions independently** through `road-to-estate-drawdown`'s consolidated decision sheet, because both read the same council record. The merge kept both, so each blocker now carries the sheet's *choice* and this branch's account of *what shipped for it*.

## Gates

All 13 exit 0, plus `task typecheck-ts` clean, **806 hook tests** passing, and `task sync` + `task generate-tools` producing no drift. `roadmap:progress` **archived nothing here** — it warns, and once 0.1–0.3 flipped it printed the Iron-Law-3 hold on the two deferred items instead. The six roadmaps that appear archived in this tree were archived on the trunk and arrived through the merge.

## Two incidents disclosed

The spend-limit stop landed between the merge resolution and its commit — nothing corrupted, and `git status` / `git log` were re-read rather than trusted from memory.

**And a self-inflicted one worth naming:** `test -f .git/MERGE_HEAD` is the **wrong probe in a worktree**, where `.git` is a *file* pointing at the real directory — so it reports "merge lost" every time. Acting on that false reading, a `git reset --hard HEAD` discarded the roadmap flips, the README row and the budget walk-down. Recovered because all four artefacts had been copied aside first, then re-applied **programmatically against the freshly merged tree** rather than restored from the copies, so main's newer content was not overwritten. `git rev-parse MERGE_HEAD` is the correct probe and confirmed the state had been intact throughout.

One caveat on history: commits `4d119cf90` and `2830bea75` came from a hunk-split that briefly left a non-compiling intermediate. The amended result typechecks and HEAD is clean, but that pair's authoring history is messier than the diff suggests.


matze4u:
**Drain-run diagnosis — moved to the end of the queue, not abandoned.**

Sync with `main` is done and pushed (`a5da0fdbc`): three conflicts resolved
(gate-violation-baselines per-key with the stricter count, estate history
append-only with main as base, dashboard regenerated). Estate walked down
61 → 55 for six blockers verified open on `main` and resolved here.

What blocks green is one gate, with one cause:

`check_source_size_budget` reports 19619 against a baseline of 19427 — **192
new**. Sole contributor: `src/scripts/hooks/dispatch_hook.ts`, 1500 → 1692.
The file sits at **exactly** 1500 on `main`, i.e. deliberately pinned at the
ceiling, so any addition reds it. Raising the baseline is what the gate itself
calls a defect, so the fix is extraction of ~192 lines.

Two cohesive candidates, both genuinely this PR's subject:

- the stdin-failure policy — `stdinReadFailure`, `_is_fail_closed_blocking`,
  `denyOnStdinFailure` (~82 lines). These are **pure**: `_stdin_read_failed` is
  written only at the call site (line 1317) and read at 1318/1368/1369, so the
  state can stay behind and only the functions move.
- the feedback/summary cluster — `_write_feedback` (132), `_record_rule_trips`
  (47) and `SUMMARY_INVOCATION_CAP`, which the diff edits heavily.

Both together clear the 192 comfortably; the stdin trio alone does not (82).

Not done in this drain pass on purpose: `dispatch_hook.ts` is the dispatcher
that runs on every tool call and carries the deny-capable guards, so a
three-module split of it is not work to squeeze between two merges in a queue
drain. Everything else on this branch is ready and the diagnosis above is the
whole remaining task.

Also worth a maintainer's eye, unrelated to the gate: the roadmap lands
**complete but unarchived** — 11/11 done with 2 unresolved `[~]` items, so
Iron Law 3 correctly refuses the archive. Item 4.2 states in its own text that
the bar is the maintainer's to pre-register, so that resolution is not an agent
call. `update_roadmap_progress --check` reds locally on exactly this; no
workflow runs that check, so it is a local signal rather than a CI blocker.

PR #1588, opened 2026-08-23
roadmap: complete repo-playbooks

matze4u:
**Partial, and the title says so.** This lands `road-to-repo-playbooks` **Phase 0 in full plus step 1.1**, and resolves the blocker that gated everything downstream. The roadmap is **not** archived and its boxes are flipped only for what is done.

## 0.1 — the negative control, and the gap is a verified negative ✅

`tests/fixtures/playbooks/mono-with-generator/` — a monorepo that already ships `turbo/generators/config.ts` (encoding barrel export, co-located test, layout), a `new:component` script, and `turbo.json` tasks with descriptions.

Asked *"add a Toast component to `@org/ui`"*, the suite routes to `react-shadcn-ui` and writes files directly. That is **verified, not recalled**:

```
$ grep -rn 'turbo gen|turbo/generators|plopfile' src/skills/ src/rules/
(no matches)
```

**Nothing** in any shipped skill or rule mentions a repository generator, so no dispatch path *can* propose one — `react-shadcn-ui` and `ui-component-architect` are the only reachable component-shaped candidates and both write files directly.

**Why that is a defect, not a preference:** the generator is the only artefact that knows this repository's answers, so a generic skill's output **compiles and is wrong** — and it fails **silently**, because the result looks like a component. Nothing signals the repository's own procedure was bypassed, which is what makes it worth a gate rather than a note.

The README also says what the fixture is *not*: `turbo` is never executed (files are read as configuration — offline, installs nothing), it endorses no tool (an integration target is not derivation-attribution), and Nx/Plop are deliberately absent.

## 0.2 — ADR-244, and the blocker it closes ✅

**`b-playbook-home-in-consumer-tree` → option (b): a sixth `Playbook` context type**, in the existing contexts directory beside the five `contexts.md` defines. AI council 2026-08-23, 2/2 convergent: *"reusing established context machinery as a sixth type minimizes new surface area versus creating a parallel structure in a new directory."* A new `agents/playbooks/` directory was refused because it creates a parallel content system with its own discovery, validation and projection surface — a cost paid on every later change to either system.

Status **proposed**, because the step says the name and class are this roadmap's proposal and the record should say so.

**Two things the ADR decides that the step did not ask for**, both because leaving them open would have cost a later step:

- **No third `recommended` grade.** A grade derived from neither the tree nor a commit is a claim with no basis, and every consumer would have to honour it for it to mean anything — the requested-rather-than-enforced label failure.
- **Nx and Plop are out of the first release.** Their discovery needs a consumer binary, while the Phase-3 staleness check must run *without* one. A gate that needs the consumer's toolchain installed is a gate that does not run.

Evidence **E2** with a claim→basis table, and it states its own limit: whether a playbook changes what an agent does is **not** measured — that is Phase 4, which pre-registers the measure with a named falsifier before Phase 2 lands.

## 1.1 — the template ✅

`src/agent-src/templates/playbook.md`. A **Source of truth** line and a **Verify** line per step, both grade obligations as a table, and two sections that earn their place: *"What this playbook does NOT cover"* (a playbook that silently stops short is worse than one that names where it ends), and the staleness paragraph stating that a failing `invokes` id **downgrades the grade** — the mechanism that keeps `configured` meaning what it says.

It opens by stating that a playbook **outranks a shipped skill**, because that is the point of the class and a template that buries it produces playbooks nobody prefers.

### A gate was right twice, and the second time was the useful one

`check_discovery_frontmatter` first read the example frontmatter as a declaration *about* the template. Moving it into a `yaml` fence fixed that — and is the better shape anyway: **every** sibling in that directory is frontmatter-less, because a template's metadata is filled in by the consuming command.

Then strict mode asked for frontmatter it no longer had. The tree already had the right mechanism, and it is not an exemption I invented: `src/config/discovery/unassigned-artefacts.yml` is a quarantine list already holding **26** template entries under the header *"They carry no runtime frontmatter on purpose: the consuming command fills it in."*

## What is left open

1.2 (`playbook-authoring` skill) · 1.3 (per-workspace `AGENTS.md` slot) · Phase 2 (precedence rule, UI lane, router hint) · Phase 3 (staleness gate + grade downgrade) · Phase 4 (pre-registered measurement).

Phase 4's ordering constraint is worth noting for whoever picks this up: its measure must be **filed before 2.2 lands**, so the precedence work cannot go first.

## Verification

`task preflight` green on the committed tree. `check_adr_frontmatter`, `generate_index --check`, `check_references`, `check_no_roadmap_refs`, `check_estate_count`, `lint_roadmap_blockers` and `build_discovery_manifest --strict` all green.


matze4u:
**Update — the roadmap is now complete and archived.** The PR was opened as a partial (Phase 0 + the template); every remaining phase landed on this branch, all six acceptance criteria are discharged, and `roadmap:progress` auto-archived the file at `count_open == 0` with zero deferred items.

## What landed after the original PR body

| Step | What | Evidence |
|---|---|---|
| 1.2 | `playbook-authoring` skill **plus** a deterministic derivation | fixture yields `add-ui-component.md` `grade=configured` `invokes=[turbo gen component]`; the same fixture yields an `observed` playbook for the unresolved `turbo gen workspace` |
| 1.3 | per-workspace `AGENTS.md` points, never restates | `findRestatedSteps`; 17/17 |
| 2.1 | `playbook-precedence` rule — matched on scope, **dispatched on grade** | § Routing; router 113 → 114 |
| 2.2 | the UI lane proposes the repository's own generator first | ordering asserted by index; empty-home half by comparing two full `StepResult`s; 10/10 |
| 2.3 | `command-routing` step 2b | 2 hits on one addition |
| 3.1/3.2 | consumer-side staleness gate | rename → fails naming file and id, restore → passes, in one test over one cloned fixture; 10/10 |
| 4.1/4.2 | pre-registration, then the measurement | verdict **confirmed** |

## Two things I got wrong and did not quietly fix

**One row of the 4.2 table was fabricated.** The first draft claimed `grep -rn 'turbo gen'` over the work engine went from 0 hits to non-zero. Running it says **0** — the lane hard-codes no vendor command at all; it reads `invokes` ids out of the repository's own playbooks at runtime, so a vendor-string grep finds nothing *by design*, and a non-zero reading would have meant the opposite of what this phase wanted. The number was stated without being run, in the direction that flattered the change. Corrected in place, with the reason, in both the evidence file and the roadmap.

**Two matrix rows and two guards were hollow passes.** In 1.3, removing the pointer carve-out left the suite green — the fixture's link label carried no step text, so nothing distinguished carve-out from no carve-out. The fixture's pointer now quotes the step in its own label, which is the hardest case and the only one that makes the carve-out observable. Same for the length guard, now exercised with a short id (`gen`) against prose containing *"generator"*. A test never seen red has unknown sensitivity, and two of these had none.

## F1's cost is published, not trimmed

The playbook path **does** read more — one directory listing, measure 3 going 1 → 2. F1 does not fire because its condition is *more reads **without** a repo-specific command*, and measure 2 went from **nothing** to the generator this repository owns. The row that got worse is in the table.

## What this does NOT establish

- Nothing measures whether a playbook improves generated code. Excluded in the pre-registration, still excluded.
- No real consumer repository was measured — one controlled fixture, one generator kind.
- Nx and Plop stay undiscovered **by decision** (their discovery needs a consumer binary and the Phase-3 check must run without one). A repo carrying either gets a partial set **with a printed notice**, never a silent one.
- The derivation script is maintainer-side: a consumer install receives the skill and not the script, so the grading rules — not the script — are the deliverable. `lint_consumer_internal_refs` caught the first draft telling them otherwise.

## Gates

Full roadmap battery green, `task preflight` clean apart from a documented worktree-only partition red (`.claude/rules` etc. duplicated against this machine's global layer — absent in CI; the branch was pushed from the checkout whose projection is partitioned). No gate skipped, weakened, or baselined around; three separate line-growth debts paid by extraction rather than a baseline bump.

PR #1590, opened 2026-08-23
roadmap: complete component-library-lifecycle

matze4u:
Phases 0 and 1 of `road-to-component-library-lifecycle` — the fixture, the pre-state, and the
library-as-a-package skill with its deterministic check. **5 of 23 steps; the roadmap stays
open** and is not archived. Phases 2–5 (Storybook, registry publishing, `DESIGN.md`
inventory, the compatibility refresh) are untouched.

## The blocker, resolved by council

`b-bundler-choice-for-fixture` was `Owner: maintainer`. AI council 2026-08-23, 2/2 quorum
(anthropic/claude-sonnet-4-5 + openai/codex-default), **convergent** on an explicit
combination: **(c)** for the source-consumed fixture plus a **static built-package-surface**
fixture modelled after **(b)**, with no bundler installed and none executed. The maintainer
delegated owner-reserved blockers to the council for this autonomous drain run; the council
decides *how*, not *whether*.

Both reviewers independently rejected the single-package alternative — a manifest declaring
both `"source": "./src/index.ts"` and `"import": "./dist/index.js"` models **one hybrid
package**, not two variants, `"source"` is a custom condition no runtime generally selects,
and the blocker's own `Resolved when` asks for the check to run against *both* variants,
which two manifests express directly.

**And both rejected the name `buildable` for the second root.** A hand-authored `dist/` is a
built-package *surface*; calling it buildable is what would have made the 1.2 test overclaim.
It ships as `built-surface/`, and the fixture README carries a § What `built-surface/` is NOT
saying plainly that a real Vite or tsup config can emit a different layout and that
establishing buildability needs a separate integration test step 1.2 cannot stand in for.

Their recorded counter-argument is kept in the roadmap rather than resolved away: the
built-surface fixture **can pass while a real bundler config fails**. Accepted, not solved.

## Step-by-step evidence

| Step | Verify as written | Outcome |
|---|---|---|
| 0.1 | `ls …/src/Button` lists both files; `jq .peerDependencies` names `react` | **discharged, at paths one level deeper** — the council's two-root layout means there is no single manifest at the fixture root. Both facts hold under `source-consumed/`; the written paths are stale, not unmet. |
| 0.2 | evidence file quotes the two skill outputs at a pinned commit | **discharged, method changed and recorded** — see below |
| 1.1 | skill + `evals/triggers.json` with two named positives; `check_references` green | **discharged**, both positives verbatim |
| 1.2 | fixture → 0 errors; `react` in `dependencies` → 1 error | **discharged** — 0 errors on both roots, exactly 1 (`peer-as-dependency`) on the mutated copy; 14/14 |
| 1.3 | § Release has three branches and cites both skills by path | **discharged** |

## Three places where the step could not be followed literally — none dropped silently

**0.2 records the skills' own text, not a transcript.** Both named skills are prose: their
"output" is whatever an agent produces after reading them, which is neither reproducible nor
re-derivable by a reader. The pre-state therefore records what they **instruct an agent to
look for**, read off the bodies at `cc1e0376b` — which is exactly where the roadmap's expected
finding ("neither names the package surface") is decidable. And **one near-miss is disclosed
rather than counted**: `exports` has 2 hits in `existing-ui-audit`, both the
component-descriptor field at `:82` and `:247` — a component's *props*, not a package's
public surface. A flat 0 would have been the cleaner-looking number and the wrong one.

**`suggested_by` is not a schema key.** `skill_linter` rejects unknown properties, so the
pack-suggestion intent is recorded in prose. React and TypeScript consumers still receive the
skill through `engineering-base`, which both packs require — the routing holds, only its
expression moved.

**The borrow-ledger entry is NOT added, and that is the decision.** The
buildable-vs-source-consumed table is derived from this fixture and from `package.json`
semantics with no external shape adopted, so under `code-provenance` there is nothing to
record; the section is labelled own analysis. A ledger row for a borrow that did not happen
makes the ledger unreliable in the direction that matters.

## What the check refuses to do, asserted rather than described

- **Classification comes from declared export targets, never a directory name.** Both fixture
  roots live under `ui-lib-vite/` and they classify **differently** — a name-based reading
  would have to call them the same thing. A *mixed* declaration is reported `undeclared`
  rather than resolved to a guess.
- **It never compiles.** A test writes syntactically broken TypeScript into the fixture's
  component and expects **zero** findings. A checker that grew a parser would report compile
  errors as packaging errors, and the two have different fixes.
- **Read-only and offline by construction** (§ Security constraints, required by
  `skill_linter` because the skill ships `scripts/`): no network, no subprocess — so running
  it over an untrusted repository executes none of that repository's code, which a package
  manager's lifecycle scripts would — and no writes, so it can never "fix" the version field
  § Release forbids touching.

**Four guards sabotage-proven:** peer placement, condition order, the mixed→undeclared rule,
and target existence. Removing any one takes the suite RED.

## Honest nulls

- **Nothing here proves any bundler can produce the declared layout.** No bundler is installed
  and none is run; that limit is stated in the fixture README, in the skill, and in the
  roadmap.
- **No real library was measured** — one two-root metadata fixture.
- **Phases 2–5 are untouched.** `count_open` is 18, so the roadmap is not archived and
  `archive_completed_roadmaps` was not run.

## Gates

Roadmap battery green (`lint_roadmap_blockers`, `check_roadmap_trackable`,
`check_references`, `check_estate_count`, `lint_framework_leakage`), `task preflight` reports
**No regressions detected**, and the two suites are 14/14 plus the check's own self-test. No
gate skipped, weakened, or baselined around.


matze4u:
**Update — the roadmap is now complete and archived.** Opened as a partial (Phases 0–1); every remaining phase landed on this branch. **22 of 23 steps, 7 of 8 acceptance criteria**, and one descope recorded as such.

## What landed after the original PR body

| Step | What | Evidence |
|---|---|---|
| 2.1 | `storybook-workshop` — one concept per story, `@summary` required, `!manifest` curation | `react-shadcn-ui` keeps a pointer whose heading still carries the word, so a reader grepping there finds the route |
| 2.2 | one a11y finding **without** a browser | `{color-contrast, story:LowContrast, error, 1.23}` from the fixture; the two passing stories in the same file yield nothing |
| 2.3 | `existing-ui-audit` § 4b — MCP precedence | live read wins, file read never removed, React-only limit with docs 10.5 cited |
| 3.1/3.2 | § Publish a registry + `checkRegistry` | both deprecated v4 types forbidden; the installer gate applies to `build` as to `add`; 22/22 |
| 4.1/4.2 | § Owned components, read first by the architect | one-row derivation from the fixture asserted; 12/12 |
| 5.2 (first half) | the Tailwind v3/v4 branch | both skills branch on the detector's `css` axis, so they agree by construction |

## The descope — and the workaround I refused

**5.1 → `stubs/road-to-scaffold-pinned-stack-versions.md`**, with **AC-8**. The step required committing the `package.json` files that come out of `npx shadcn@latest init` and `npx storybook@latest init`. `shadcn@4.19.0` is **interactive past the flags the step names**: a project-name prompt, then a monorepo prompt, then an **arrow-key component-library select** a piped newline cannot answer. Three attempts — the N=3 budget for one target — then stop instead of guessing further at flags.

`npm view shadcn version` answers in one call and returns `4.19.0`. **Writing that into § Compatibility was refused on purpose:** a registry lookup is not a scaffold, and the step's rule is that the major came out of a *working* scaffold committed beside it. A better-sourced wrong-shaped claim is still the defect that roadmap exists to correct. AC-8 is descoped with 5.1 rather than ticked on a weaker basis.

The useful half holds regardless — **no skill touched here states a shadcn or Storybook version at all**, so none states one it cannot point at. The stub carries a three-way promotion criterion.

## Two defects of mine, recorded rather than quietly fixed

**A hand-edit of a generated page.** Fixing count drift I rewrote a dated measurement in `docs/proof.md` so a 2026-08-12 figure would not be silently rescaled. The reasoning was right and the *layer* was wrong: `build_proof` renders that page from `docs/CLAIMS.md`, so the edit broke the drift guard and the next regeneration would have discarded it anyway — and `update_counts` already anchors that figure in the source, which makes rescaling it the maintainer's established choice, not mine to override from the projection.

**An estate-ratchet red that was not a real addition.** `active_roadmaps` read 9 → 10 because this branch predated #1588's merge, so the archived playbooks roadmap counted twice across the two branches. Resolved by merging `main`; the ratchet was never touched.

## Honest nulls, restated at closure

- **No bundler was installed or run.** Nothing here proves the declared layout is producible — which is exactly why the second fixture root is named `built-surface` and not `buildable`, a naming both council reviewers insisted on.
- **The browser a11y path never ran.** What ships is a declared-colour contrast check that cannot see a rendered page, and a colour behind a token indirection is invisible to it. Asserted, not merely stated.
- **No real consumer library was measured.** One controlled fixture, one generator kind.
- **The two prose skills** whose behaviour steps 0.2 and 4.1 asked about were verified by *what they instruct*, not by a transcript — a transcript proves a model's behaviour, not the instruction's correctness.

## Gates

Full roadmap battery green — blockers, complexity, ci-steps, trackable, no-roadmap-refs, empty, later-disposition, estate-count, references, source-size, depth-budget — plus `task preflight` clean. `check_artefact_count_messaging` fired a second time when the two new skills moved the count to 293 and was fixed by regenerating `CAPABILITIES.yaml` and `docs/proof.md` from their sources. **No gate skipped, weakened, or baselined around at any point.**

PR #1689, opened 2026-08-27
roadmap: complete undeclared-obligation-disposition (the ceiling admits no declaration)

matze4u:
Closes `road-to-undeclared-obligation-disposition` — fourteen steps, six acceptance criteria, one blocker. The roadmap archives in this branch.

## The finding

**82 of 120 rules declared no enforcement at all.** Two independent readings agree on the same tree — `grep -L 'enforced_by\|instruction-only' src/rules/*.md` and the instrument's own `undeclared` — so the cohort is a fact about the tree, not an artefact of one parse. The instrument had been printing that number, green, for months. It moved 85 → 82 by accident over that time; **this is the first time it moved by decision.**

```
before:  declared 38 · local-only 0 · observer 10 · unwired 0 · missing 0 · undeclared 82
after:   declared 52 · local-only 3 · observer 14 · unwired 0 · missing 0 · undeclared 68
```

**The headline did not move, and that is correct.** 15/120 rules have a backstop that fails a CI build, before and after. Nothing here made an obligation safer at runtime — risk 3 of the roadmap's register names exactly that trap, and the instrument's own class separation is what prevents the change from reading as coverage.

`unwired 0` and `missing 0` held across the change: every declaration resolved to a carrier that exists and is reachable. **The instrument performed every downgrade itself** — four `hook:` declarations became `observer` because their manifest entries are `fail_closed: false`; three `validator:` declarations became `local-only` because no workflow invokes `task ci`, which `consistency.yml:159` states in as many words. Nothing self-graded.

## The buckets

| bucket | count | landed |
|---|---|---|
| already-carried | 13 | **7** |
| cheaply-probeable | 56 | no, by design — Phase 2 decides these |
| instruction-only-by-nature | **7** | **7** |
| structurally-blocked (kernel) | 6 | no — a guard denies the write |
| **sum** | **82** | 14 |

**56 is the number to be suspicious of, and it is reported rather than trimmed.** Risk 1 warns that a bucket which swallows everything converts an honest unknown into a false settled state. But the dangerous bucket is `instruction-only-by-nature` — that one writes dispositions — and it is **7**. The large bucket is the one that changes nothing until Phase 2 ranks it.

## Six were NOT declared, and that is the finding rather than an omission

A declaration resolving to `unwired` is a **defect class**, not a neutral record. Four rules have a plausible carrier that **nothing runs**:

| rule | carrier | reachable from |
|---|---|---|
| `history-discipline`, `scale-discipline` | `lint_persistence.ts` | **nothing** |
| `runtime-safety` | `lint_skill_frontmatter_safety.ts` | **nothing** |
| `cross-source-consistency` | `bench_cross_source_eval.ts` | **nothing** |

**Three linters exist in `src/scripts/` and no workflow, taskfile or config invokes any of them.** Two more rules had a plausible but unverified carrier. Over-assigning `already-carried` is the single failure that would make this exercise false, so the bar was "I read the carrier and it enforces the obligation".

## The one probe, and the ranking that earned it

Of the 56 probeable rules, **exactly one** carries a cited measurement of its own failure: `user-interaction.md:75` — *every* malformed ask was an unblocked one, and **no gate ships** for that class because `check_reply_consistency` reads "exactly the surface that did not fail". **1 with a measurement, 55 without** — the direction 2.1 predicted, by 55×.

`src/scripts/probe_unblocked_ask.ts` measures what that checker structurally cannot:

```
  hand-back turns          117
  …with a numbered block   36   (excluded — check_reply_consistency reads that surface)
  …unblocked asks           4   3.4% of hand-backs
  …of those, malformed      4   100.0% of unblocked asks
```

**All four are the `sag Bescheid, wenn …` shape the 2026-08-06 audit named.** The probe reproduces that audit's finding independently, on a different corpus, without being told what to look for. A **ceiling**, not a point estimate, and it prints both bounds itself. `--self-test`: 7 cases, 3 positive, 4 negative — the negatives are load-bearing, since an instrument that fires on everything measures nothing. **Stopped at one** (2.3).

## The ratchet — and a defect it exposed in its own instrument

Blocker `is-a-declaration-worth-anything` resolved **(a)**, decided from the distribution as its own recommendation (c) asked. A ratchet gates a constant only if most of the cohort is unobservable; **7 of 82 are**. `check_enforcement_coverage --check` now guards `undeclared` shrink-only — sabotage-verified: removing one declaration gives `undeclared rules rose: 68 → 69`, exit 1; restored, exit 0.

**Landing Phase 1.3 fired two of its existing checks, and neither event happened:**

```
· frequency gaps rose: 9 → 14 (a rule's carrier no longer fires often enough…)
· validators fell back to taskfile-only: 0 → 3 (a gate left .github/workflows/…)
```

A rule with no declaration has no carrier, so it contributes to **neither** counter — which means **declaring a rule truthfully can only raise both**. Both compare against a baseline taken when the rule was invisible. The instrument punishes honest declaration in proportion to how much honesty a change lands. The baseline is regenerated — the response its own failure message names — and the reasoning is written into the script beside the new check, because *a regenerated baseline and a hidden regression are indistinguishable in a diff*.

## 221 tokens, a zero-headroom ceiling, and a council that did not converge

The 14 declarations cost **221 tokens** of per-spawn preamble payload. `check_preamble_payload_budget` blocks in CI against a grace ceiling of **138,212** whose config says it **may never move UP** — and `origin/main` measured **exactly 138,212, the ceiling to the token**. So the constraint is not specific to this change: *any* rule-metadata addition is currently blocked.

Every reason was compressed to a clause and the full reasoning moved to the evidence report, which is not in the payload. That recovered **227 of the original 448** and is the "transfer" the prior lock (*"rule growth is transferred, never funded"*) asks for. The remaining 221 is structural.

**AI council split 1–1:**

| seat | verdict |
|---|---|
| `anthropic/claude-sonnet-4-5` | ship, record 221 as transfer debt with a numeric repayment target, no ceiling raise |
| `openai/codex-default` | find ≥ 221 tokens of headroom in the **same change** — "transfer debt" turns a hard invariant into a queue of promises |

Both agreed the change is worth making and that **all 14 should land**; the dissenting seat argued explicitly against dropping the seven `instruction-only` ones, because they distinguish *consciously unbacked* rules from rules nobody assessed.

**What landed is the debt option**, with the dissent kept rather than resolved away and the offset attempt reported as **failed**: every candidate was drive-by editing of prose the change does not touch, or editing the gate. The dissenting seat's own warning applies to its alternative — same-change savings "may incentivize rushed compression that makes important instructions worse".

The concurring seat's binding conditions are discharged in full: **no ceiling raise** (138,212, expiry 2026-11-10 unchanged), a **numeric** repayment target, and the debt recorded where a maintainer reads it — `stubs/road-to-preamble-transfer-debt-221.md`, counted in the dashboard's owner-decision total, not inside an archived roadmap.

## Three corrections to the plan, recorded rather than applied silently

- **AC-5 asks for nine kernel rules; the list is six.** Three of the nine already declare enforcement and were never in the cohort. And **two of the six would NOT be `instruction-only`**: `commit-policy` and `scope-control` are both carried by `block_unauthorized_git`. The guard is suppressing two *true declarations*, not nine unknowns.
- **2.1 cites `session-canary` as a second measured rule.** It is not in this cohort — it already declares enforcement. A valid example, not a second candidate.
- **Phase 4's `gate-coverage.yml` registration is NOT landed**, and says so: the instrument emits no `scanned:` line, and the manifest is being edited by open PR #1682 — the step's own text says to coordinate with it. The ratchet, which is 4.1's substance, is landed and proven.

## One dead end, recorded so it is not retried

A YAML flow declaration (`enforced_by: ["…"]`) would cost one line instead of two. The frontmatter parser does not read it — the declaration resolved to `missing 1`.

## Verification

- `check_enforcement_coverage` — `undeclared 82 → 68`, `unwired 0`, `missing 0`; `--check` exit 0, sabotage-verified
- `probe_unblocked_ask --self-test` — 7/7 (3 positive, 4 negative)
- `lint_regression` — no regressions (the `long_rule` warning it first raised is paid for inside the same file, no clause removed)
- green: `lint_roadmap_blockers`, `lint_roadmap_complexity`, `check_roadmap_trackable`, `check_no_roadmap_refs`, `lint_empty_roadmaps`, `lint_plan_risk_register`, `check_references`, `lint_evidence_artifacts`, `lint_canonical_terms`; `tsc --noEmit` clean


matze4u:
**The PR body above describes what was measured; this comment describes what ships, because CI changed the answer.**

Three reds, and reading them was worth more than the original change.

## The declarations do not ship — none of them

`tests/scripts/check_preamble_payload_budget.test.ts` refuses them. The grace ceiling is enforced by a **test**, not only by the report-only workflow — so "ship it and record the debt", the option a split council landed on, **was never available**. The framing that led the council to it was mine.

| tree | tokens | vs ceiling |
|---|---|---|
| `origin/main` | **138,212** | exactly on it, to the token |
| + all 14 declarations | 138,433 | +221 |
| + the 7 carrier declarations only | 138,284 | **+72** |
| + none | 138,195 | **−17** |

**Even seven do not fit.** No rule may gain an `enforced_by` field at all while main sits on the ceiling — a repo-wide constraint this cohort merely discovered.

All 14 are reverted and none is lost: the evidence report carries each with its **exact declaration string** (§ *The 14, ready to apply*), and the stub is reframed from "repay 221 tokens" to "apply these when there is headroom". Applying them later is mechanical.

## Two of them paid off anyway — and that gain ships

`tests/scripts/rule_backstops_ci_wired.test.ts` refused two `validator:` declarations because **no workflow ran** `lint_legal_pack` or `lint_roadmap_ci_steps`. The test names the fix in as many words, so both are now in `.github/workflows/rule-backstops.yml`.

**Two linters that nothing ran now run in CI. The enforcement headline moves 15/120 → 17/120, for real.** Declaring the rules is what surfaced it.

## What the roadmap now records

**AC-2 is marked NOT MET** rather than reworded to fit what shipped. The 82 → 68 drop is real, reproducible from the report, and not in this tree.

This is the **second framing error** of the run — the first was the adoption-floor question on #1687, also mine, also reversed on the corrected record. Both are recorded in the roadmap, the report, the stub and the run summary rather than quietly dropped. Twice in one run is a pattern, and the pattern is mine: I write the framings.

## Still shipping

- the full disposition of all 82 rules into four buckets, sums closing
- `probe_unblocked_ask` — 4 unblocked asks in 117 hand-backs, **all 4 malformed**, every one the shape the 2026-08-06 audit named
- the `undeclared` ratchet, shrink-only, sabotage-verified (baseline back to 82)
- the finding that the enforcement ratchet **punishes honest declaration**
- the finding that **three linters exist and nothing invoked them** — two now wired

## One local false red, not fixed because it is not a defect

`explain_run`'s two "honest no-data" cases fail in any worktree where a PostToolUse hook has written `agents/state/context-hygiene.json`. Gitignored, absent in CI, and the tests pass in two sibling worktrees at the same commit.

PR #1692, opened 2026-08-27
roadmap: complete executable-specification-layer

matze4u:
Completes `road-to-executable-specification-layer`. Phase 1 shipped in full; Phases 2 and 3 descoped by AI council into a stub with written promotion conditions. Every open step is now `[x]` or `[-]`; the roadmap is archived in this branch.

## What shipped

**Phase 1.1/1.2 — the routing decision the suite did not have.**
`src/skills/test-case-discovery/SKILL.md` gains `## Does this change owe an executable behavior contract?`, containing:

- **The observable-behavior test** — does this alter behavior someone outside the code can observe, *and* would that person describe it in their own words? Both halves, because observable behavior with no outside audience is still only a case.
- **Four cases answering no**, against a floor of two: a refactor with no behavior change; an internal invariant with no external vocabulary (cache keys, retry backoff, query plans); a pure boundary or error path with no business reading; a change whose only observer is another part of this system.
- **The anti-script rule** as a fenced Iron Law, with a worked wrong-then-right pair — the same cart/coupon behavior written once as five interface steps (`When I click #apply-coupon`, `And I wait 2 seconds`, `Then .cart-total should contain "€45.00"`) that break on a redesign and survive a moved discount rule, and once as three sentences about money. Closed by a portability test: could the step still be true if the product were a CLI, an API and a web page?

**Phase 1.3 — the rubric stops asserting a convention taught nowhere.**
`src/skills/judge-artifact-completeness/rubrics/ticket-quality-score.json` → `acceptance_criteria.criterion` now routes the Given-When-Then question to that section and grades independently-testable outcomes; `gap_signal` gained the anti-script case, so 1.2's rule has a consumer inside the judge on the day it lands.

**Blocker `grading-a-dimension-nothing-teaches` — implemented, not merely recorded.**
`src/scripts/grade_target_readiness.ts:177-208`: `test-strength` → `advanced-testing-signals`, `grade: null`, new `observations: string[]` reporting `mutation-testing-config-detected`, `property-testing-library-detected` and `mutation-testing-ci-enforcement-detected` independently, and a `notDetectable` scoped to test *effectiveness* rather than to the dimension — the config is detectable, what it implies is not — emitted whether or not an observation fires.

Three epistemic states are held apart on purpose: signal observed, no signal observed, effectiveness not evaluable. An empty `observations` array is the second; no `neither-detected` string is ever emitted, which would re-create the verdict the unscore removed.

## The measurements this rests on, re-run at HEAD

| Measurement | Result |
|---|---|
| `grep -rniI 'gherkin\|cucumber\|behat\|given.when.then' src/` before the change | 1 hit — `ticket-quality-score.json:21`, a rubric asking a *reviewer* for the shape |
| `grep -c 'observable behavior' src/skills/test-case-discovery/SKILL.md` | 2, one inside the added section |
| `grep -c 'click #' src/skills/test-case-discovery/SKILL.md` | 1, and it is the negative example |
| `grade(full).dimensions` id set | no longer contains `test-strength` (asserted) |

## Gate resolutions, with the mechanism for each

| Gate / blocker | Mechanism |
|---|---|
| `which-stack-gets-the-first-adapter` | AI council 2026-08-27, anthropic + openai, **2/2 convergent** → **(c)**. Recorded in the roadmap's Blockers entry with the dissenting argument and the revisit-if. |
| `grading-a-dimension-nothing-teaches` | AI council, **split in round 1**, **2/2 convergent in round 2** → unscore + rename. Implemented, not just recorded. |
| Phase 3 disposition | AI council round 2, mechanism-match test → **same mechanism** as the sibling roadmap's measured refusal → whole phase descoped. |
| `lint_roadmap_blockers` | ✅ 9 roadmaps blocker-contract-clean, decidability 0 violations |
| `lint_roadmap_complexity` | ✅ 9 complexity-clean |
| `check_roadmap_trackable` | ✅ 6 active, all parseable |
| `check_no_roadmap_refs` | ✅ 980 scanned, no roadmap references in stable artifacts |
| `lint_empty_roadmaps` · `lint_roadmap_later_disposition` | ✅ |
| `check_estate_count` | ✅ within ratchet — `+0 active / -1 disposed` |
| `lint_plan_risk_register` | ✅ 9 ready roadmaps clean |
| `check_condensation` · `check_references` · `lint_output_slop` · `check_md_language` | ✅ |
| `lint_canonical_terms` | ✅ 995 violations against baseline 1007 — unchanged by this branch, not lowered on a local reading |
| `task typecheck-ts` · `eslint` on changed TS | ✅ exit 0, no findings |
| `npx vitest run` | 18178 passed / 1 failed — see below |

## The one red, attributed

`tests/scripts/check_rule_projection_integrity.test.ts:198` — `expected 13 to be greater than 50`.

Not this diff: `git diff --name-only origin/main...HEAD | grep rules/` is empty. It is the known worktree-environment red — `.agent-settings.yml` is absent from the worktree, so the ADR-236 delivery partition projects only the project-layer rules and the plan is *partial* rather than empty, which is the case the test's own `it.skipIf(planEmpty)` guard does not cover. Measured green on all eight CI shards previously. CI is the authoritative reading.

## Sensitivity — the new assertions were probed before being trusted

Backup-copy restore each time, never `git checkout`:

| Sabotage | Result |
|---|---|
| Restore `grade: _graded(mutation \|\| property, mutationCi)` | **4 red** of 27 |
| Emit `neither-detected` when nothing is found | **1 red** of 27 |
| Restored | **27 pass**, `diff` against the backup empty |

The eight new specs also assert both probe directions (`python/` has `[tool.mutmut]` and no mutation config in `full/`; `full/` has `fast-check` and `python/` does not), the empty-observation case and the CI-enforcement case against purpose-built temp targets — the shipped fixtures cannot express either, so both branches would otherwise be uncovered.

## Council decisions, in full

**Round 1 — `which-stack-gets-the-first-adapter`.** Options: (a) TypeScript via `playwright-bdd`, (b) Laravel via Behat, (c) neither first. Verdict **(c)**, both seats. Both named (b) as the better guess when a request arrives — the suite's PHP depth is real and the Pest boundary is a sentence rather than a design. One seat raised the only argument against: the existing rubric line is itself weak evidence that the pattern matters, so implicit demand may already exist — resolved by making that demand the *trigger* rather than the warrant. The other seat explicitly declined a request-count trigger, on the grounds that demand intensity, maintenance cost and fit are the axes, not arithmetic. Revisit-if: a consumer names a stack **and** a concrete workflow, or recurring production-validation findings show a stack-specific gap prose cannot close.

**Round 1 — `grading-a-dimension-nothing-teaches`: split.** Seat A for (a)-with-a-label-fix, seat B for (b)-withdraw. Seat A conceded seat B's factual point that config presence is not a reliable strength measure. Not resolved on the split.

**Round 2 — the split, put back.** Both seats converged on **unscore + rename to `advanced-testing-signals`**, and both specified the same shape independently (`grade: null`, an `observations` array, `notDetectable` scoped to effectiveness). Seat A's proposed numeric reopen thresholds (">20 survivors", ">2hrs") were rejected by seat B as unsupported by any record here and are **not adopted** — noted in the roadmap so a later reader does not inherit an invented cutoff.

**Round 2 — Phase 3, against a measured refusal.** Mechanism-match run first: **same mechanism**. A changed-surface adapter differs from the refused rig in *selection scope*, not in mechanism — both generate mutants and classify survivors and timeouts. The archived refusal in `archive/road-to-test-independence-and-mutation-evidence.md` is measured (10 probes in minutes, 3 survivors, hand-probing kept up) and its own reopen condition — a survivor population too large to hand-probe — has not fired. A `revisit_if` describes what would lift a `degraded` state; it is not authorization to build past a later, evidence-backed refusal. Both seats: **(b)**, descope all of Phase 3. Seat A's supporting inference — a narrower surface must yield fewer survivors — was rejected by seat B as not generally reliable; the verdict is unaffected and the correction is recorded so the weaker argument is not inherited.

**Also decided:** Phase 2.1 (BDD-runner detection) ships in neither branch. It was not formally blocked, but both seats found its only named consumer to be the adapter (c) declined.

## Two corrections recorded rather than quietly applied

1. **The roadmap's own measurement table understated the dimension.** It named only the mutation probe; the code also detected property-based testing (`fast-check`, `hypothesis`), which round 1 did not discuss. So the grade was mixing two signals under one label. Corrected in the table, marked superseded.
2. **Step 1.1's verify string asked for `behaviour`.** `src/config/canonical-terms.yml:79-80` makes `behavior` canonical and `behaviour` a ratcheted variant, currently at 995 against a baseline of 1007. Satisfying the literal would have added violations to a tightening gate in order to pass a verify. The canonical spelling was used and the correction written at the step; the verify's intent — a named, greppable discriminator — is met.

## Honest nulls

- **No registry `state` was flipped.** `mutation-sensitivity` stays `degraded`, `e2e-test` stays `unknown`. That is the property AC-6 exists to protect, and it holds even though AC-6 is itself descoped. What changed in the registry is descriptive: two `projection` renames, one added `limitations` line recording the unscore, and a `revisit_if` qualified so it can no longer be read as authorization.
- **AC-3 through AC-6 are `[-]`, not `[x]`.** Each depends on a descoped phase, and each is carried verbatim into the stub so promotion restores them rather than re-deriving them.

## Descope

`agents/roadmaps/stubs/road-to-executable-specification-adapter.md` holds Phase 2 (2.1–2.3), Phase 3 (3.1–3.3) and AC-3–AC-6, each with why it is held and what would promote it. Its § 5 carries the parent's kill register forward unchanged: no Acceptance IR, no parser, no generic adapter API, no external toolchain on latest-upstream pins, no Gherkin-for-every-change, no acceptance-green as a trust signal, no universal minimum mutation score.


matze4u:
Merged `origin/main` (#1691) — that PR edited this roadmap in the same window and the conflict is resolved keeping **both** sides:

- **Step 2.1** — #1691's council (2/2) caught a real contradiction in the original wording ("never a second framework" while naming Behat alongside Pest; `playwright-bdd` is an added dependency too unless already installed). Its corrected two-branch text is kept **verbatim**, with this PR's `[-]` descope applying on top. A step being descoped is no reason to archive a version of it that is known to be wrong, and `preservation-guard` binds on the passage either way.
- **Acceptance criteria** — #1691's framing is adopted: AC-1/AC-2 are the closure set, AC-3–AC-6 were conditional on the demand blocker. It resolved **(c)**, so those four close `[-]` with the blocker as the reason. Its AC-4 correction is also kept: the original required the uncovered count to equal the whole detection list, which is arithmetically impossible once one adapter ships.

The two changes never disagreed on substance — neither ships an adapter now.

Re-verified after the merge: `lint_roadmap_blockers`, `lint_roadmap_complexity`, `check_roadmap_trackable`, `lint_plan_risk_register`, `check_estate_count`, `check_references` all green; projections regenerated.

matze4u:
**Neutral review, and it found three defects in this PR's code. All three are fixed** (commit `8d23352b7`).

A stop-gate flagged that this session mutated 118 non-doc lines with no neutral review. A cross-model review was commissioned over the **complete** non-doc delta, split mechanically by file group; two seats, neither having written the code or the prompt's expectations.

1. **`mutation-testing-ci-enforcement-detected` overclaimed.** Static matching over a workflow file proves a mutation tool is *referenced*; it cannot prove the step is enabled, blocking, reached on the required branches, or ever executed. One seat argued the `notDetectable` disclaimer covers it; the other answered that a consumer is entitled to read a token's name. Renamed to `mutation-testing-ci-reference-detected` — the probe is unchanged, only the claim it makes. Recorded as a `limitations` line in the registry.

2. **The level spec was deletion-insensitive.** It asserted `grade(full).level > 0`, which stays green if the dimension is **deleted outright**, so it proved nothing about the mechanism its own comment claimed. Both seats named it. Rewritten to assert the relationship: the dimension exists, is `null`, is not a knockout, and the level equals an independently-computed minimum over knockouts only.

3. **`evidence` and `observations` were two representations of one fact.** The ternary reported only the mutation signal when both fired — which the `python` fixture does — so the human-readable string and the machine list disagreed about the same target. `evidence` is now derived from `observations`.

**Sensitivity, probed after each fix:**

| Sabotage | Red |
|---|---|
| delete the dimension entirely | **8** |
| fold non-knockouts into the level | **4** |
| restore the parallel ternary | **2** |
| restored | 29/29, tree clean |

**Worth stating plainly: the first version of fix 3 had no spec behind it and its sabotage passed green** — the same defect class the review had just named, one commit later. Two specs were added and the sabotage now reds.

The id rename was checked rather than assumed: `grep -rn "test-strength"` returns the two registry `projection` fields and the one test, all updated here, and no other consumer. One seat correctly noted the rename is a breaking contract change for any caller doing id lookup — there are none in this tree.

The three findings are recorded at the blocker in the archived roadmap, so a later reader sees them where the decision lives rather than only in a PR comment.

PR #1693, opened 2026-08-27
roadmap: complete composition-before-creation

matze4u:
Completes `road-to-composition-before-creation`. Every step is `[x]` except 4.2, which is `[~]` and **carried** into a follow-up created in this same change; the roadmap is archived here. Closing it required a small tooling change to the archival sweep — see § The sweep change.

## What the estate never recorded

`artifact-drafting-protocol` Phase B has mandated a four-surface overlap scan since it shipped, and ships `instruction-only`. The scan leaves no artifact, so a skill authored after a thorough one and a skill authored after none are indiscernible in the tree. That is why the three overlap tools (`audit_overlap`, `audit_skill_overlap`, `report_layer_overlap`) run over a *corpus* rather than over a *decision*.

**Phase 1 measured it** — the twenty most recently added skills and rules by first-commit date, 2026-08-06 to 2026-08-23. Detector validated **in both directions before the count was read**, which step 1.3 requires: known-positive `src/rules/fix-what-you-see.md:77-87` (`## Why separate from active-remediation`) → positive; known-negative `src/skills/experiment-loop/SKILL.md` (no incumbent named anywhere) → negative.

The step's own premise turned out wrong in a useful way — "record" admits three readings:

| Reading | Count | Kill criterion (≥15/20) |
|---|---|---|
| Machine-readable (what the Goal actually asks for) | **1 / 20** | does not fire |
| Strict prose — an authoring *decision* about a named incumbent | **9 / 20** | does not fire |
| Loose prose — a `See also` gloss or a `Do NOT use when` row | **18 / 20** | *would* fire |

So no single rate could decide the shape. Both council seats rejected the loose reading and **continued**: a routing row names a neighbour without recording that it was evaluated as an incumbent or why creation stayed justified. Full table with `file:line` per positive and an explicit blank per negative: `agents/evidence/analysis/authoring-search-record-sample-2026-08-27.md`.

## The finding that shaped the mechanism

A machine-readable "named incumbent + one-line delta" field already exists and is already enforced: `collision_ok` (`rule.schema.json:195`, gated by `lint_trigger_collisions.ts:146`), carried by **47 of 120 rules** and **0 of 299 skills**. Two properties keep it from being what was needed — it is keyed by *trigger value*, not incumbent, and it fires only on a collision, so substantive overlap without a shared trigger is invisible. What it does establish: authors here **do** fill in a one-line delta field when a lint asks, at 39% adoption with a hard gate behind it.

## What shipped

**`composition_review:`** in both artifact schemas (15 lines each, inserted rather than round-tripped, so the diff carries no reformat). Named against `skill.schema.json`, never against the source's list — `requires` is untouched, per the roadmap's own collision note.

```yaml
composition_review:
  - candidate: skill:existing-ui-audit     # skill: / rule: / command: / guideline: / none
    disposition: create_separate
    rationale: audits what exists before a write; this governs the write itself
```

Four dispositions. `create_separate` is the value **no incumbent vocabulary in this tree can express** — an incumbent exists, was evaluated, and a separate artifact was authored anyway. The nearest incumbent is the harvest ledger's `adopt | adapt`, and it supplies the argument: `lint_harvest_provenance.ts:222` deliberately **excludes** rejection because a rejected harvest has no artifact to cite, while this record is written *on* the artifact being created, so rejection is what it most needs. `none_found` exists so an author is never pushed to invent an incumbent to satisfy the schema — that failure would manufacture the pro-forma record the mechanism is supposed to measure, using the mechanism.

**`lint_composition_review`**, with its teeth split deliberately:

- **Hard (exit 1)** — a present record whose `candidate` resolves to nothing, whose `none`/`none_found` pair disagrees, whose disposition is off-enum, or which repeats a candidate. A record naming an incumbent that is not in the tree reads as evidence of a search that cannot have happened, which is worse than carrying no record.
- **Advisory (exit 0)** — an artifact *added* relative to a base ref carrying no record. Reported by path.

The corpus is the whole estate so the coverage floor can be honest (419 live, 380 floor); `lint_evidence_artifacts` declines registration for exactly the reason a diff-scoped gate cannot carry one. The diff narrows only the advisory half, and its addition set unions `git ls-files --others --exclude-standard` so a create-only canary is visible.

**Registered across all six surfaces**: `gate-coverage.yml` (with a `canary:` recipe), `taskfiles/ci-fast.yml`, the `ci:` list in `Taskfile.yml`, and a `consistency.yml` step with identical argv and exit-2-degraded handling. Header denominators updated **in place, same line count**, so the `.secret-allow` PEM pin held at line 487.

## The sweep change — why closing this roadmap needed one

Iron Law 3 says a `[~]` deferral must be resolved before archival, and the preservation test routes four dispositions to the council. Three of them — carry to a follow-up, merge into existing work, restore to open — **leave the glyph in place**, while `archive_completed_roadmaps.ts:336` required `deferred === 0` unconditionally. So a genuinely **unexecutable** item, one needing elapsed time rather than effort, had no council-reachable path to an archived roadmap: the only glyph that cleared the gate was `[-]`, which is owner-reserved *and* pinned to "won't happen at all". Using it for a carried item makes cancellation and transfer indistinguishable to every later reader.

A `[~]` step may now carry `<!-- deferred-resolution: carried-to=<slug> -->` inside its own block (`merged-into=` for a merge into pre-existing work). `[-]` stays cancellation-only. **Validation fails closed at every step**, which is the condition both seats attached to their verdict:

| Door | Verdict |
|---|---|
| bare `[~]` | blocks |
| malformed annotation | blocks (not read as absent) |
| destination does not exist | blocks |
| destination in `archive/` or `skipped/` | blocks — a dead roadmap cannot receive an item |
| `carried-to` destination with no `parent_roadmap` back-link | blocks |
| back-link naming a *different* parent | blocks |
| `merged-into` destination that never mentions the source | blocks |

**Not checked, and stated rather than implied:** that the destination's copy of the criterion is faithful. Both seats asked for stable item identity over verbatim text matching, and neither exists today — the back-link plus the source mention is the strongest link the sweep can verify without inventing an identity scheme.

## Verification

| Check | Result |
|---|---|
| `lint_composition_review` | ✅ 419 artifacts, floor 380 |
| `--self-test` | ✅ 8/8 behaved, 5 rejecting |
| `check_gate_coverage --canary` | ✅ *"caught the planted contract-violation defect (exit 1)"* |
| advisory half, probed separately | ✅ untracked artifact reported by path at **exit 0**, tree clean after |
| `check_gate_coverage` · `check_ci_local_parity` | ✅ every floor cleared · 153 CI / 290 local |
| `gate-self-test:registered-non-adopters` | unchanged at 24 — the new gate is an adopter |
| `validate_frontmatter` | ✅ 450 artifacts, 0 failing, 0 warnings — the optional field costs 299 edits of zero |
| roadmap battery (blockers · complexity · trackable · no-refs · empty · later) | ✅ |
| `check_estate_count` | ✅ +1 active / −1 disposed, 1 exempt |
| `lint_plan_risk_register` · `check_condensation` · `check_references` | ✅ |
| `lint_canonical_terms` | ✅ **992**, three *below* where the branch started |
| `check_rule_stub_ceiling` · `check_always_budget` | ✅ (ceiling raised with a reason — below) |
| `task typecheck-ts` · `eslint` on changed TS | ✅ exit 0 |
| `npx vitest run` | 18199 passed / 4 failed — all four attributed below |

## Sensitivity — every new assertion was probed

Backup-copy restore each time, never `git checkout`:

| Sabotage | Result |
|---|---|
| harden the presence half into a violation | **3 red** of 17 |
| drop the `none`/`none_found` cross-check | **2 red** |
| add `deferred` to the disposition enum | **2 red**, naming the review vocabulary it collides with |
| accept a bare `[~]` in the sweep | **2 red** of 15 |
| accept a dead destination | **2 red** |
| disable the step-block boundary | **1 red** |
| restored | 17/17 and 15/15, `diff` empty |

**One of those probes caught a defect in my own test.** The block-boundary spec was written with the annotation *above* the unannotated step, where the assertion holds whether or not the boundary exists — it passed against a build with `BLOCK_END_RE` disabled. Corrected to put the annotation after, and the sabotage now reds. The uncorrected version is the exact shape of a test that never proves anything.

## The four reds, each attributed

1. **`check_rule_projection_integrity`** — the known worktree red. `.agent-settings.yml` is absent from a worktree, so the ADR-236 delivery partition projects a *partial* plan, which the spec's own `skipIf(planEmpty)` guard does not cover. No rule files in this diff. Green on all eight CI shards previously.
2. **`check_preamble_payload_budget`** — same root cause, settled by arithmetic: 138,277 tokens against a 107,646 ceiling, dominated by `project-scope rules 122,680`. This branch adds **+74**; a +30,631 overshoot is not attributable to it.
3. **& 4. `explain_run` ×2** — a real latent defect, reproduced both directions: **19/19 green** with the gitignored `agents/state/` moved aside, **2 failed | 17 passed** with it present. `baseOpts()` pins three of four inputs into its fixture and leaves `hygiene: null`, which falls back to `DEFAULT_HYGIENE_CANDIDATES` — relative paths resolved against the repo root. Not fixed on sight because a second spec writes that file *inside* the fixture and also passes `null`, so it passes today for the wrong reason; the two specs want opposite things from the same `null`, and settling it is a question about the script. Tracked in `stubs/road-to-hermetic-explain-run-tests.md` with the reproduction, the cause, four closing steps and a sibling sweep — per `fix-what-you-see`, a red is fixed with its verification or gets a tracked follow-up, never named and handed back.

**No ceiling and no baseline was lowered to make anything green.**

## Council decisions

**Q1 — which reading governs the kill criterion.** Both seats: it **does not fire**, continue. They split on *which* reading (strict 9/20 vs machine-readable 1/20) and agreed on the outcome; both explicitly rejected the loose 18/20 reading. One seat argued the criterion tests whether the practice is "already organic"; the other that the Goal's explicit "machine-readable" wording governs the ambiguous later phrase. Recorded because the reading cannot be relitigated after implementation.

**Q2 — extend `collision_ok` or add a field.** Both **(b)**, a new sibling field. `collision_ok` represents permission to share a *trigger*; re-keying it would alter a CI-enforced contract 47 rules already populate while losing its trigger-specific meaning. One seat specified the multi-candidate shape and the explicit "searched, found nothing" value, both adopted.

**Q3 — `disposition-vocabulary-authority`.** Both **(b)**, its own enum, with (c) unification deferred rather than rejected: it reaches into release tooling before a shared domain model has been shown to exist.

**Q4 — the carried-deferral glyph.** Both **(c)** over an overloaded `[-]` or a permanently-active roadmap, with fail-closed validation as an attached condition.

**Rejected and recorded:** one seat proposed numeric thresholds (75% / 60% vocabulary overlap, 20% adoption, 6-month windows, ">20 survivors", ">2hrs"); the other rejected them as unsupported by any measurement here. **None is adopted**, so a later reader does not inherit an invented cutoff.

## Corrections recorded rather than quietly applied

1. **Step 2.1 undercounted the vocabularies — six, not four.** It missed `lint_harvest_provenance.ts:76` (`adopt | adapt`), which is the closest incumbent of all and supplies the argument for the fifth enum, and `build_archive_index.ts:71-76`'s `Disposition`. All six are pinned with their `file:line` in the disjointness suite.
2. **Step 2.3's venue was wrong.** `canonical-terms.yml` maps *spelling variants*; an enum value is not a misspelling of anything, so registering there would have satisfied the verify literally while guarding nothing. Scanned count is 1,586 before and after — unchanged, and that *is* the finding. The obligation moved to a disjointness suite whose sensitivity is proven.
3. **Step 3.3 cost a ceiling raise, stated at the step.** `check_rule_stub_ceiling` refused twice; after two shrink passes (928 → 783 → 754 tokens) the residual +74 is obligation surface that cannot move without the rule mandating a field it does not name. Raised **680 → 754 — the exact measured count, no slack** — with a `history` entry. **Hand-set rather than `--write-baseline`**, because that mode also re-anchored five unrelated ceilings downward (`context-hygiene` 2470→1979, `improve-before-implement` 1385→840, and three more): real shrink-only gains, but ones this change neither made nor measured, and locking them in here would be a ratchet moved on a local reading.
4. **This branch's own prose used the ratcheted `artefact` spelling**, taking `lint_canonical_terms` 995 → 1006 — under baseline, so green, but adding violations to a tightening gate. Normalized to 992. The first attempt was a silent no-op: BSD `sed` has no `\b`, so the substitution matched nothing and exited 0.

## 4.2 — carried, not dropped

`road-to-composition-review-false-positive-rate.md`, created in this change: `status: ready`, `parent_roadmap` back-link, a `> Blocked until …` note, step 4.2 reproduced **verbatim** with its phase context. It writes the true-positive / false-positive definitions **before** any collection (so the definition cannot be shaped by the numbers), names the two known false-positive shapes with one anchored to a real artifact, requires the population to be greater than zero — a rate over an empty population closes nothing — and forbids proposing a flip itself. Estate: +1 active, −1 disposed, no blocker.

## Merge note

`origin/main` advanced to `460b62007` (#1691) mid-branch; merged in, projections regenerated, every gate above re-run after the merge. That PR also edits `road-to-executable-specification-layer.md`, which my open PR #1692 archives — flagged there, not here.


matze4u:
**Correction, and it was mine.** The `check_preamble_payload_budget` red on shard 3/4 was **not** a worktree artifact — I attributed it wrongly in the PR body and CI was right.

I compared against the **design** ceiling (107,646), saw a +30,631 overshoot, and concluded a +74-token rule edit could not be the cause. The operative ceiling is `ci_delivery.grace_ceiling` = **138,212**, set to the exact measured total on 2026-08-24 with **zero slack** and expiring 2026-11-10 — the same no-slack pattern I applied to the stub ceiling in this very branch, which is what makes the miss avoidable rather than unlucky.

Measured properly, in a clean worktree of `origin/main` versus this branch:

| Ref | Measured | vs ceiling 138,212 |
|---|---|---|
| `origin/main` | 138,195 | −17 |
| this branch, before the fix | 138,277 | **+65** |
| this branch, after the fix | **138,208** | −4 |

So the branch had **17 tokens** of headroom for the whole project-scope rule payload, and the addition was 74.

**Fixed by shrinking, not by raising.** Raising a grace ceiling whose entire purpose is to force the number down by November would be a ratchet moved for my convenience. The five-line paragraph is gone; `composition_review:` is now named inside the existing Phase-B parenthetical — cheaper, and better placed, because the field is named where the phase it belongs to is described. Step 3.3's obligation (the rule cites the field) still holds.

`check_rule_stub_ceiling` corrected in the same commit: the rule now measures **690** body tokens, not 754, so the ceiling and its history entry both move to 690. Leaving it at 754 would have described a change that no longer exists.

The other three reds in the PR body stand as attributed — `check_rule_projection_integrity` (worktree projection, no rule files in the diff) and the two `explain_run` specs (reproduced both directions against the gitignored `agents/state/`, tracked in `stubs/road-to-hermetic-explain-run-tests.md`).

matze4u:
**Neutral review, and it found nine defects in this PR's code. All nine are fixed.**

A stop-gate flagged that this session mutated 118 non-doc lines with no neutral review. A cross-model review was then commissioned over the **complete** non-doc delta — selected by `git diff`, split into four parts **mechanically by file group** because it exceeded the 51,200-byte transport ceiling, no file excluded. Two seats per round; neither wrote the code or the prompt's expectations.

### `lint_composition_review` — five

1. **The contract over-claimed.** The docstring and **both** schema descriptions said a lint checks that `candidate` resolves, while `command:` and `guideline:` were silently exempt — `command:this-does-not-exist` passed everything. `artefactIds` now walks all four trees; commands resolve below their pack, guidelines without the extension.
2. **`composition_review: []` was accepted** — present, saying nothing, neither of the two states the gate distinguishes.
3. **A git failure read as "no additions".** An unresolvable base ref exited 0 with zero advisories: a blind run indistinguishable from a clean one, which is the exact failure `gate-coverage.yml`'s own header describes. Now throws `GitScopeError`.
4. **The advisory path had zero coverage**, despite a comment calling its `ls-files --others` union load-bearing. Real-git fixture now.
5. **Two parser holes** — a YAML block scalar captured `|` and produced a one-character value; the candidate grammar admitted `guideline:/foo`, `foo/` and `foo//bar`.

### `archive_completed_roadmaps` — four

6. **The back-link slug was interpolated into a `RegExp` raw.** `road.parent` matched `roadXparent`, and a slug containing `[` made `new RegExp` **throw and abort the sweep** — inside a function whose entire contract is failing closed.
7. **A roadmap could name itself** as destination: passes every check, then dies with its own archival.
8. **A destination could go dead in the SAME sweep** — the sharpest finding. Two roadmaps complete, parent carries to child, both validated live, both archived, carried item left with no receiver. The mechanism meant to prevent the loss produced it.
9. **`merged-into` was satisfied by a substring** — a filename, an example or a comment counted as a link. Structured now: a `relates:` row or a `parent_roadmap:` line.

Plus the gap both seats named independently: **every unit spec stayed green if `deferralProblems()` were deleted from `archive_completed()`**, so nothing proved the validation was wired. Three integration specs now drive the real CLI over a real git repo.

### Claim accuracy

The gate said a record pointing at a non-existent incumbent "reads as evidence of a search that cannot have happened" — in the docstring, the operator message and the manifest note. That is a claim about the **search**, and the gate has none to make. What it certifies is **referential consistency**. Narrowed in all three places, with the roadmap's own Risk 1 named as what this does not close.

### Sensitivity

| Sabotage | Red |
|---|---|
| restore the skill/rule-only carve-out | 2 |
| accept an empty record | 1 |
| git-failure-as-empty | 1 |
| loosen the candidate grammar | 1 |
| stop marking block scalars | 1 |
| unwire `deferralProblems()` from the sweep | 2 |
| unescape the slug | 1 |
| substring `merged-into` | 1 |
| drop the same-sweep guard | 1 |
| drop the self-reference guard | 1 |
| restored | 24/24 and 21/21, tree clean |

### Three of my own failures, same class as what the review found

- **Two specs had to be corrected rather than kept.** One asserted the very carve-out that turned out to BE the defect. One measured the coverage floor against `artefactIds` instead of the scanned corpus, so it reddened the moment candidate resolution widened — while the floor it checked had not changed.
- **A fix with no spec behind it.** The first version of the `evidence`-derivation fix passed its own sabotage green, one commit after the review had named exactly that class.

### Contamination reverted — please note this one

`git checkout -B` carries uncommitted changes across branches, and a later `git add -A` committed three ERD artifacts into this PR:

- `road-to-database-erd-landing.md` with all phases `[x]` and both blockers resolved — which would have marked that roadmap **complete on main with none of its work landed**, and made the archival sweep archive it;
- its Phase-3 evidence file;
- a `schema-erd` **skill-admission row** — admitting a skill present in neither PR.

All three restored to `origin/main`. The evidence file moved to #1694 where its roadmap lives; the admission row is in **neither**, because it belongs with the landing.

PR #1694, opened 2026-08-27
docs(evidence): the estate/payload ratchet collision that blocks the ERD landing

matze4u:
`road-to-database-erd-landing` was executed end to end and **could not land**. This PR ships what the attempt produced: a measured, reproducible finding that two governance ratchets disagree about whether the estate may grow, and the roadmap updated so the next attempt starts from that measurement rather than from scratch.

**The 5,456 tested lines are not in this PR.** They are safe on `feat/schema-erd-diff` and on the local `drain/database-erd-landing` branch, cherry-picked clean onto current `main` with original authorship preserved. Nothing was discarded.

## The finding

| Ref | Measured | vs grace ceiling 138,212 |
|---|---|---|
| `origin/main` (460b62007) | **138,195** | −17 |
| + the `schema-erd` skill, full description | **138,248** | **+36** |
| + the same skill, description gutted to 12 tokens | **138,213** | **+1** |

The third row is the finding. The description was cut from 198 characters to 56 — destroying the `'show me the schema'` trigger — and the total is **still one token over**, because the catalog bucket is `Σ "- <name>: <description>\n"` (`preamble_byte_census.ts:399`) and `"- schema-erd: "` alone measures **6 tokens** before a single word of description.

**So no skill of any description can land while the grace ceiling sits at the exact measured total.** That is a property of a zero-slack ceiling, not of this skill.

Meanwhile `check_estate_count` sets `skill_count` allowance 0 and names the escape explicitly — *"An addition takes the `estate_growth_exempt` claim path — a recorded reason read from the diff — or it fails"* — and a prior council round granted exactly that exemption for this skill. One gate says growth is allowed if justified; the other says growth is not allowed and *"It may never move UP"*. Both block.

## Why every escape is closed

| Escape | Why |
|---|---|
| Raise the grace ceiling | Forbidden by its own config text, unconditionally. Its purpose is convergence toward the design ceiling by 2026-11-10; raising it claim-by-claim is convergence in reverse. |
| Shorten the new skill's description | Measured insufficient by construction (row 3), and it destroys the trigger the skill's reachability argument rests on. |
| Shorten other skills' descriptions | Lowering a ratchet on a local reading of surfaces this change does not own. Both council seats forbade it — it funds an addition by shaving unrelated semantics and punishes whoever moves first. |
| Reduce the census formatting | The catalog representation is **synthetic**; the host's real on-wire format is not file-measurable (`:400`). Editing it lowers the number by redefining the metric — worse than raising the ceiling, because it hides the change. |
| Ship without the skill | Refused 2/2 on reachability: `grep -rlE 'erDiagram' src/skills/*/SKILL.md` returned **0 of 299**, and an implementation agents cannot discover does not deliver the capability. |

## What the execution actually verified

Every phase ran against a real cherry-picked branch. None of this is carried from the branch's own archived roadmap, which is the artifact Phase 2 existed to distrust:

| Step | Evidence |
|---|---|
| 1.1 | 33 files, 5456 insertions, 25 `A` entries — re-measured against current `main`, not the 2026-08-26 reading |
| 1.2 | 8 conflicts, **zero** under `src/scripts/schema_erd/`, `src/skills/schema-erd/` or `tests/`. A ninth surfaced during the real cherry-pick that `merge-tree` does not report: `agents/roadmaps-progress.md`, now gitignored |
| 1.4 | Divergence list **empty**, and the step's premise was wrong: `ir.ts` shares no type with `_lib/persistence/` — that directory is a *finding* pipeline for `lint_persistence`, so "the current adapter shape" names a contract these surfaces never had in common |
| 2.1 | `git merge-base origin/main HEAD` == `git rev-parse origin/main`, confirmed **before** any generator ran |
| 2.2 | **The step earned its place.** `README.md` was *not* wholly generated — `HEAD` carries a hand-authored "How these are counted" block the branch's side lacks. A blanket "take the branch's regenerated version" would have deleted it silently. All eight hunks were stale counts (291/117/200/106 vs 299/120/202/116) |
| 2.3 | `task sync` then `generate-tools`; `check_condensation` green; counts re-derived to 300 with the hand-authored block intact |
| 2.4 | Fresh run: `validate_frontmatter` 451/0/0 · `check_references` ✅ · `check_claims` ✅ · `check_source_size_budget` at baseline · `typecheck-ts` 0 · the branch's own **134/134** |
| 2.5 | Two sabotage probes, named: guarding `prisma.ts:203`'s `foreignKeys.push` → `prisma adapter > emits the foreign key on the owning side only`; removing `mermaid.ts`'s `tokens.push('FK')` → `renderMermaidDiff — markers are the primary channel > emits PK, FK and UK key tokens`. Restored 134/134, `diff` empty |
| 2.6 | Admission recorded. **The step's framing was corrected**: it asks for `why_not_extend` against "the 34 skills in `family: backend-data`", but there is **no `family:` frontmatter field** in this tree — `grep -rl '^family:' src/skills/*/SKILL.md` returns nothing. Nine adjacent skills enumerated by hand instead |
| 3.1 | `agents/evidence/analysis/unlanded-finished-branch-2026-08-27.md` — a **seven**-day gap, not six, and the mechanism named: closeout 7.2's verify joins two clauses with *and* that have different observability, so the checkable half passed and the conjunction reported its value |
| 3.2 | **No probe, and the count is the reason**: 1,146 local branches, **193** with unmerged commits. One mattered. The discriminator that actually worked is recorded for a future probe: "unmerged commits **and** its own roadmap archived as fully closed" |

## Council decisions

**Landing shape** — cherry-pick over rebase, and **all five commits** rather than four. Both seats chose cherry-pick (it leaves `feat/schema-erd-diff` untouched, so no inherited commit is rewritten or dropped). They split on the count; the wider reading was taken because it satisfies both seats' concerns and neither's objection survives it — authorship and dates preserved, the archived roadmap reaches `main` as a real record rather than a re-typed one, no manual re-commit, and the deleted `INDEX.md`/`index.json` are not resurrected. Seat 1's own reservation — that omitting commit five is "a substantive selection decision" — is the argument *for* carrying it.

**Skill vs artifact** — the branch's opt-in experimental shape lands, 2/2, on the measurement rather than by inheritance. Two alternatives named and rejected: a generated artifact with no skill (three inbox proposals argued it; refused on reachability) and a default-on skill (nobody proposed it, and it is the routing problem the first alternative feared arriving by the other door).

**Estate allowance** — **(a) refined**, 2/2, and both seats rejected the roadmap's own recommendation of (b). This roadmap takes **one** exemption; the `-relational-modeling` sibling's is **not** pre-authorised. Approving two prospectively converts a deliberate zero-growth ratchet into unlimited-if-plausible; the allowance's text names `estate_growth_exempt` as *the mechanism*, not permission to use it freely. Seat 1 corrected seat 0's framing and the correction is adopted: the sibling gets a **fresh evaluation**, never a predetermined "must extend".

**The ratchet collision** — **(a)**, 2/2: keep both gates independent, the ceiling holds, block until a separate payload-reduction change creates headroom. Both rejected raising the ceiling and both rejected excluding the catalog bucket (that would stop detecting verbose descriptions and cumulative growth, which `check_estate_count` cannot see). One seat disputed my framing and the correction is adopted: the gates govern **different scarce resources**, so this is not a logical contradiction — but the policy demands an offset while defining **no owned reduction mechanism**, which makes the freeze real in practice. Its words: *"a budget with no ownership model or reduction mechanism can become deadlock dressed as process."*

**November milestone** — both seats: **not known-unmeetable, but known-unplanned and not credible.** 30,549 tokens over the design ceiling, no mechanism, no per-bucket targets, no owner. Recorded in the evidence file rather than in the config, because changing a registered budget's milestone is a maintainer decision this analysis does not take.

**Rejected and recorded:** numeric thresholds proposed by one seat (5% / 10% cumulative growth caps, a six-month revisit trigger) were rejected by the other as unsupported and are **not** adopted.

## Verification

`lint_roadmap_blockers` ✅ · `lint_roadmap_complexity` ✅ · `check_roadmap_trackable` ✅ · `check_no_roadmap_refs` ✅ · `lint_empty_roadmaps` ✅ · `lint_roadmap_later_disposition` ✅ · `lint_plan_risk_register` ✅ · `check_condensation` ✅ · `check_references` ✅ · `lint_output_slop` ✅ · `lint_evidence_artifacts --new-only` ✅ · `task typecheck-ts` exit 0 · `check_estate_count` ✅ (`open_blockers` 43→44, authorised by the claim).

`check_preamble_payload_budget` reports `138195 > 107646` locally — that is the **design** ceiling and the number is **exactly `origin/main`'s**, so this branch adds zero payload. CI's `standing-payload-delta.yml` applies the grace ceiling.

## What this PR does not do

- **Proposes no reduction.** Where 30,549 tokens should come from is a separate question. 89% of the payload is the `project-scope rules` bucket (122,680 tok), and nothing here says that corpus is too large — only that nothing currently walks it down.
- **Changes no ceiling and no baseline.**
- **Does not claim the grace ceiling was set wrongly.** Setting it at the measurement was a deliberate, recorded council decision (2026-08-24, 2/2). This records a consequence that decision did not anticipate.
- **AC-1 is back to `[ ]`.** It is the one criterion that cannot be met yet, and marking it done would be the silent-green defect this repository has recorded before.

### DIFF
--- a/agents/evidence/analysis/estate-payload-ratchet-collision-2026-08-27.md
+++ b/agents/evidence/analysis/estate-payload-ratchet-collision-2026-08-27.md
@@ -1,0 +1,111 @@
+<!-- evidence-type: analysis -->
+# Two ratchets disagree about whether the estate may grow, and one of them makes any new skill unlandable
+
+**Measured 2026-08-27** while attempting `road-to-database-erd-landing`. The
+finding is about the **gate configuration**, not about that roadmap: it applies
+to every skill addition, and it was discovered by trying one.
+
+## The two gates
+
+**`check_estate_count`** — `skill_count` allowance is 0, and
+`src/config/estate-count-budget.json:24` names the escape explicitly:
+
+> "0 — no allowance, deliberately. … An addition takes the
+> `estate_growth_exempt` claim path — a recorded reason read from the diff — or
+> it fails."
+
+So growth **is** permitted, conditional on a recorded justification.
+
+**`check_preamble_payload_budget`** — `src/config/preamble-payload-budget.json`
+→ `ci_delivery.why_a_grace_ceiling`:
+
+> "The grace ceiling is set AT the measurement so growth beyond today reds
+> immediately while today's tree passes, and it ratchets DOWN toward the design
+> ceiling as reductions land. **It may never move UP.** It expires at the
+> milestone-1 date, at which point the design ceiling applies."
+
+`grace_ceiling: 138212`, `grace_measured_at: 2026-08-24`,
+`grace_end_date: 2026-11-10`, `design_ceiling: 107646`.
+
+So growth is **not** permitted, and the ceiling has no claim path.
+
+## The measurement
+
+Clean worktrees, `check_preamble_payload_budget` on each:
+
+| Ref | Measured | vs grace ceiling 138,212 |
+|---|---|---|
+| `origin/main` (460b62007) | **138,195** | −17 |
+| + the `schema-erd` skill, full description | **138,248** | **+36** |
+| + the same skill, description gutted to 12 tokens | **138,213** | **+1** |
+
+**The third row is the finding.** The description was cut from 198 characters to
+`"ERD of a schema, or a diff-ERD of what a change alters."` — 56 characters,
+destroying the `'show me the schema'` trigger phrase — and the total is *still*
+one token over. The floor is structural: the catalog bucket is
+`Σ "- <name>: <description>\n"` per skill (`preamble_byte_census.ts:399`), and
+`"- schema-erd: "` alone measures **6 tokens** before a single word of
+description. With 17 tokens of headroom, name plus formatting plus any usable
+description does not fit.
+
+**So no skill of any description can land while the grace ceiling sits at the
+exact measured total.** Not a property of this skill; a property of the ceiling
+having zero slack.
+
+## Why the obvious escapes are closed
+
+| Escape | Why it is closed |
+|---|---|
+| Raise the grace ceiling | The config says "It may never move UP", unconditionally. Its whole purpose is to force convergence toward the design ceiling by 2026-11-10; raising it claim-by-claim is convergence in reverse. |
+| Shorten the new skill's description | Measured above: insufficient by construction, and it destroys the trigger the skill's reachability argument rests on. |
+| Shorten other skills' descriptions | Lowering a ratchet on a local reading of surfaces the change does not own. Both council seats forbade it explicitly — it funds an addition by shaving unrelated semantics and punishes whoever moves first. |
+| Reduce the census's own formatting | The catalog representation is **synthetic** — the census constructs `"- name: description\n"`; the host's real on-wire format is not file-measurable (`:400`). Editing it lowers the number by redefining the metric, which is worse than raising the ceiling because it hides the change. |
+| Ship the capability with no skill | Refused 2/2 on reachability: `grep -rlE 'erDiagram' src/skills/*/SKILL.md` returned 0 of 299, and an implementation agents cannot discover does not deliver the capability. |
+
+## Council verdict, 2026-08-27
+
+AI council, 2 seats (anthropic + openai), **convergent on (a)**: keep both gates
+independent, the ceiling holds, and a new preloaded skill must satisfy the
+payload ceiling on its own — an estate exemption authorises **count** growth
+only, never every resource the skill consumes.
+
+Both seats rejected raising the ceiling, and both rejected excluding the catalog
+bucket from it (that would stop detecting verbose descriptions and cumulative
+growth across individually-approved skills, which `check_estate_count` cannot
+see).
+
+One seat disputed the framing: the gates govern **different scarce resources**, so
+"a PR satisfying one necessarily fails the other" is overstated — a new skill can
+satisfy both *if its payload is offset*. That correction is adopted, and it
+sharpens rather than dissolves the finding: the policy demands an offset while
+defining **no owned reduction mechanism**, which makes the freeze real in practice
+even though it is not a logical contradiction. Its words: *"a budget with no
+ownership model or reduction mechanism can become deadlock dressed as process."*
+
+Both seats' fallback, and the disposition taken here: **block the addition until
+a separate payload-reduction change creates headroom.**
+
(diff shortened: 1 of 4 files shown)


matze4u:
**Neutral review context.** A stop-gate flagged that this session mutated 118 non-doc lines with no neutral review, and it was right. A cross-model review was commissioned afterwards over the **complete** non-doc delta — split mechanically by file group, no file excluded, two seats per round. Twelve findings, all real, all fixed; nine of them landed in #1693 and three in #1692. **This PR carries no code**, so none of the twelve are here — but two corrections to this PR's own artifacts came out of the same cleanup, and both matter for how it reads:

**1. Cross-branch contamination, reverted from #1693 and partly moved here.** `git checkout -B` carries uncommitted changes across branches, and a later `git add -A` had committed three ERD artifacts into #1693 — where none of them belongs:

- this roadmap with all phases `[x]` and both blockers resolved, which would have marked it **complete on main with none of its work landed** and made the archival sweep archive it;
- its Phase-3 evidence file;
- a `schema-erd` **skill-admission row**, admitting a skill present in neither PR.

All three restored to `origin/main` in #1693. `agents/evidence/analysis/unlanded-finished-branch-2026-08-27.md` moved **here**, where its roadmap lives, so step 3.1 is `[x]` in this PR and nowhere else. The admission row is deliberately in **neither** PR — a row admitting an absent skill is a claim nothing backs, so it goes with the landing.

**2. A wording correction in this roadmap.** It said "AC-1 goes back to `[ ]`", which claims a transition that never happened on this branch: AC-1 was never `[x]` here. It is simply the criterion that names the obstacle. An explicit note now says why **every other** step and AC stays `[ ]` — the Phase 1–3 verification all ran and its evidence is in the blockers, the Notes and two analysis files, but none of it *landed*, so marking a step done would claim a state the trunk does not have. Step 3.1 is the sole exception, because its entire deliverable is an evidence file and that file is here.

`agents/evidence/drain-run-summary.md` in this PR now carries the full review record — all twelve findings, the sensitivity probes, the contamination, and three failures of my own of the same class the review was finding (two specs corrected rather than kept, one of which asserted the very carve-out that turned out to be the defect; and a fix whose first version had no spec behind it and passed its own sabotage green, one commit after the review named exactly that).

PR #1764, opened 2026-08-31
roadmap: governed-harness Phases 4-6 (9 steps), AC audit, and the Phase 7 split (council 2/2)

matze4u:
Advances `road-to-governed-harness-evolution` from 24/59 to 33/59 done, closes two acceptance criteria, and splits the owner-blocked Phase 7 into its own ACTIVE roadmap so seven specified steps and AC-9 stay in the governed estate instead of holding the parent hostage.

## Steps closed (9)

| Step | What landed | Where |
|---|---|---|
| 4.2 | Metric **vector**, never a weighted total; `artifact-count-delta` row refused-if-missing | `src/scripts/_lib/evaluation_vector.ts` |
| 4.3 | Verdict hierarchy — `paired_verdict` per metric decides, `underpowered` is not a pass, the Pareto frontier only ORDERS | same module (`promotionVerdict:230`, `paretoFrontier:205`) |
| 4.5 | Minimality tie-break, E5's four criteria `tokens → artifacts → scope → precedence` | `src/scripts/_lib/minimality_tiebreak.ts` |
| 4.6 | Regression selection from the affected neighbourhood, with a refusal on an unresolved surface | `src/scripts/_lib/regression_neighbourhood.ts` |
| 4.7 | Evaluator promotion refused with no cross-grade; a control-satisfied plant is a discrimination deficit, never a win | `src/scripts/_lib/evaluator_promotion.ts` |
| 5.2 | No-live-harness constraint, scanned in both halves | `tests/scripts/governed_harness_no_live_harness.test.ts` |
| 5.3 | Three role prompts, three input sets, an **outcome-blind judge** in three independently-failing layers | `src/scripts/_lib/role_split.ts` |
| 5.5 | E6's seven curator ops, near-duplicate caught with zero model calls | `src/scripts/_lib/curator_ops.ts` |
| 6.2 | Single matcher preserved; parity test green plus a second-matcher detector | `tests/scripts/single_matcher_preserved.test.ts` |

**Step 4.5 also reconciled a contradiction.** Its text said *"include the fifth criterion … simpler mechanism"* while citing E5, which records a 2/2 council verdict that the fifth criterion is **OUT**. The text is amended to match, marked explicitly as a transcription of the existing verdict that creates no new decision and does not reopen E5.

### Sensitivity proved, not asserted

Every guard was neutralised, watched go red, and restored:

- **4.2** — adding `weightedTotal()` to the module: `AssertionError: expected { …(1) } to deeply equal {}` / `+ "src/scripts/_lib/evaluation_vector.ts": Array [ "weighted", "vector-to-number" ]`, `1 failed | 17 passed`. Removed → 18/18.
- **4.6** — `affected(graph, t, depth)` → `affected(graph, t, 0)`: `expected [] to deeply equal [ 'reg-neighbour' ]`, `5 failed | 8 passed`. Restored → 13/13.
- **5.3** — layer-2 surplus-key check neutralised: `1 failed | 20 passed` → restored 21/21.

## Acceptance criteria — audited, two closed, seven left open with the exact missing conjunct

| AC | Verdict | Why |
|---|---|---|
| **AC-1** | **closed** — after being fixed | Three modules had **no inventory-matrix row at all**. Rows added; conjunct 2 measured. |
| **AC-2** | **closed** | Full-estate sweep, not the 0.2 spot checks: exactly two files overlap, both already dispositioned. |
| AC-3 | open | `evaluated` still unmet — all four Phase-4 modules have **no production caller**; `evolution_lab`'s verbs have no evaluation path. |
| AC-5 | open | Conjunct 2 met **as a named-construct static scan**, not a semantic proof. Conjunct 1 not met: `promotionVerdict` has no caller, so no promotion is decided by `paired_verdict` because none is decided. |
| AC-6 | open — **reproduced defect** | Conjunct 2 met. Conjunct 1 **falsified**: the pinned `SET-SHA256 7e091dfc…` does not recompute — the artefact's own recipe yields `0667fbd9…`. Cause located: the freeze commit `34318f7f` **also edited three of the files it was freezing**; 97/100 rows reproduce. The ordering half is intact. |
| AC-7 | open | 6.1 not done; no three-arm measurement exists. |
| AC-8 | open | Conjunct 1 met vacuously (no candidate run has happened); the ROI figure is 5.6, open. |
| AC-9 | open | Transferred — see below. |
| AC-10 | open — **falsified as written** | The `no-runtime-daemon` marker in `README.md` is gone and `docs/CLAIMS.md` moved `backed` → `withdrawn`, removed by `68463a1e` under ADR-249 — a different roadmap. Left open rather than re-keyed: re-keying would close it by redefinition. |

## Council decision — Phase 7 disposition (anthropic + openai, 2/2 convergent)

**Verdict: Option 3 — split at the phase boundary into a new ACTIVE roadmap.** Both seats explicitly **rejected** parking it in `agents/roadmaps/later/`, on the ground that `later/` is excluded from the dashboard and from `/roadmap:process-*` and therefore does not preserve active-estate membership — the exact property `roadmap-progress-sync` Iron Law 3's preservation test requires.

`agents/roadmaps/road-to-harness-promotion-bridge.md` carries, byte-verified verbatim: steps 7.1-7.7, step 0.8, AC-9, the `merge-authority` blocker whole, and the Phase 0 carried condition. The parent's nine items become `[-] **MERGED (outcome transferred)**`, each stating in as many words that this means **TRANSFERRED, not cancelled and not satisfied**.

Both seats' exact texts are reproduced in the receiver: the Resume condition (which routes owner **refusal** back to the owner rather than pre-authorising a weakening of AC-9), the Revisit-if, and the Carried blocking condition — *"A check over a population of zero does not discharge this condition."*

Merge authority itself was **not** decided here and could not be: ADR-239 § Decision 3 reserves it to the owner, and the record says why — *"an agent that both wants the capability and writes the amendment authorising it is the shape the reservation exists for."*

## Findings worth reading

1. **The `### blocker:` prefix bit in the reverse direction.** The parent-side stub first read `### blocker: merge-authority — MOVED`, which `lint_roadmap_blockers` parsed as a **live** entry with id `merge-authority — MOVED` and no fields, and `check_estate_count` counted `open_blockers 31 → 32` — one blocker with two live owners, which atomic transfer forbids. Fixed by **deliberately dropping** the literal prefix on the stub, with the distinction written in so a later reader does not "repair" it back. The parent separately documents a 2026-08-29 repair of the *opposite* defect.

2. **The carried condition's discharge point may already have passed — and this run did not adjudicate it.** Parent steps 3.4 and 3.6 are `[x]`, and `src/scripts/evolution_lab.ts:858-888` carries a `promote` verb returning `EXIT_REFUSED` unconditionally, printing the lifecycle gate plus the blocker. That is a non-empty population with a mechanical refusal. Whether it **discharges** the condition is a decision the verdict does not make, so the condition transferred unchanged and the observation is recorded as explicitly not adjudicated. **This is the one thing worth a maintainer's attention.**

3. **`estate_growth_exempt` is present and honest but authorised nothing here.** `check_estate_count.ts:731` and `:765-780` set `allowance.active_roadmaps = offsets.exempt.length`, so the offset claim alone raised the allowance and `over` was empty before claims were consulted. A code-read, not a measurement.

4. **The 4.6 graph substitution is real and narrows what the step delivers.** `agent-config code-graph detect` → `no code-graph source detected` (the native engine ships `hooks.code_graph.enabled: false`), so 4.6 is built on `discovery_graph` — measured live at 785 nodes / 1672 edges. `requires_skills` is not one of its five extracted edge kinds, so a code-symbol candidate gets a weak neighbourhood. The module **refuses** an unresolved surface rather than silently selecting zero regressions, but refusing is not the same as selecting correctly. Stated in the module header, not hidden.

5. **Two stated-not-measured defaults**, both with a `revisit-if`: the artifact-delta ceiling of `0` (4.2) and the 70 % near-duplicate threshold (5.5). Neither is a measured optimum.

## Not attempted, with the reason

4.1 and 4.4 went to a separate AI council for their design gaps (the twelve cascade stages are decided in arity but enumerated nowhere; the `WHERE × WHY` vocabularies do not exist in the tree). 6.1 is spend-bearing through `description_route_check.ts:112-125` and no provider quota was spent by this branch. 6.3/6.4/6.5 depend on 6.1. Phase 7 is owner-blocked.

## Gates

`lint_roadmap_blockers` · `lint_roadmap_complexity` · `lint_roadmap_ci_steps` · `check_roadmap_trackable` · `check_no_roadmap_refs` · `lint_empty_roadmaps` · `lint_roadmap_later_disposition` · `check_estate_count` · `check_source_size_budget` · `lint_evidence_artifacts` · `lint_plan_risk_register` · `check_references` · `check_council_references` · `check_condensation` · `check_md_language` · `npm run typecheck` · `eslint` — **all exit 0**. Tests: 116/116 across the nine branch suites, plus `harness_evolution_guard_call_sites` 15/15.

Estate: `active_roadmaps 3 → 4` with both `estate_offset_exempt` and `estate_growth_exempt` claimed on the receiver; `open_blockers` unchanged at 31 because `merge-authority` moves rather than multiplies.

`task ci` was not run locally — the repo's own notes record it red on `main` without a branch-vs-base diff. Remote CI is the authoritative gate.


matze4u:
## Update — 6.1 closed on a measured three-arm run, and AC-7 with it

Since the PR body was written, two more commits landed.

**Step 6.1 — CLOSED, outcome (A), zero provider calls.** The three delivery arms turned out to be measurable deterministically, because they differ on exactly one observable property of the tree: *is the labelled rule's body in context for this prompt*. Two consumers, both read rather than recalled — `src/scripts/condense.ts:1124` (`writesThinFiles`, projection time) and `src/scripts/hooks/rule_inject_hook.ts:230` (`deliversBodies`, prompt time).

| arm | delivery | false context | standing tok | injected/prompt |
|---|---|---|---|---|
| `eager-all` | 1.000 (305/305) | 1.000 (194/194) | 120,743 | 0 |
| `thin` | 0.000 (0/305) | 0.000 | 18,223 | 0 |
| `delivery` | 0.990 (302/305) | 0.000 | 18,223 | mean 2,026 · p90 4,144 |

Evidence: `agents/evidence/analysis/governed-harness-three-arm-delivery.md`. Reproduce with `./scripts-run src/scripts/model_rule_injection --three-arm` (byte-identical across runs). The three `delivery` losses are **named individually** rather than summarised: matcher misses on `roadmap-progress-sync` and `user-interrupt-priority`, one cap drop on `think-before-action`. `thin`'s 0.000 is stated as a corpus∩arm property — no labelled rule is kernel or triggerless — not as a claim that `thin` delivers nothing.

**Why the existing price grid was not the experiment.** `model_rule_injection.ts:454-462` already priced the same three shapes. A price grid reports what a shape **costs** and never what it **delivers**; the delivery half is what 6.1 asked for and what was missing.

**Sensitivity, proved by a cap sweep:** `delivery` moves 0.770 → 0.774 → 0.990 → 0.993 as the cap goes 1 B → 2 kB → 20,480 B (shipped) → 200 kB, landing exactly on the matcher's own 0.993 recall ceiling when the cap stops binding — while **both standing arms do not move at all**, which is the other half of the proof. A second handle (stripping a rule's triggers in an in-memory router) moves its positives from the injected set into the thin standing set. Both pinned in `tests/scripts/_lib/delivery_arm_experiment.test.ts` (11/11), which also asserts the three arms report three *different* numbers.

**AC-7 — CLOSED, both conjuncts.** The second was already met; the first closes on this run; and the *"before any new retrieval component"* half is now satisfied by **ordering** rather than vacuously — 6.3 is still open. The prior audit note is preserved under the closure rather than deleted.

**Steps 4.1 and 4.4 — a council answered and returned `REVISE`, not a greenlight.** Both stay open, and both now carry a named prerequisite where before they had nothing.

- **4.1 was a DEGRADED 1/2 round.** The anthropic seat returned `exit_1` with empty stderr, 0 tokens and ~87 s latency on **three** attempts — the two-decision form twice and a split single-decision form once — while answering the other five questions of this run normally. The failure is question-shaped, not quota. Under the N=3 budget the run stopped retrying and took the single seat, per the mandate to degrade and record. A second, independent reason not to close on it: **the two openai passes produced two materially different twelve-stage enumerations**, which is what non-convergence looks like.
- **4.4 was 2/2 present and 2/2 on `REVISE`.** Both seats named the same architecturally prior gap — *"best intervention within a cell"* is undefined, so "archive the best per cell" names no operation — and both flagged a **defect in the question itself**: it grounded `WHERE` in "the six-rung ladder" and called the anchor obvious **without enumerating its six values**. The next attempt must paste `LADDER_RUNGS` into the brief. Both also ruled that a best-per-cell archive cannot *be* the diversity signal, because retaining one winner per cell erases whether attempts ran 50/50 or 99.9/0.1.

**What the answers do agree on**, and it is usable: no stage may be SOFT (a condition whose failure does not block "actually done" is a diagnostic on the receipt, not a cascade gate); `underpowered` is a hard non-pass meaning *collect evidence*; and the first failed stage maps onto Phase 1's existing four-value family **without** inventing a fifth. What blocks the rest is an *independent, append-only receipt producer* that does not exist — a predicate being implementable is not evidence that trustworthy evidence for it can be produced.

**Not verified, stated plainly:** behavioural equivalence between a delivered body and a standing one is **not** a budget deferral — ADR-202 records that instrument CLOSED (κ 0.472 against a registered 0.800 floor, no third attempt licensed). No amount of spend closes it, so there is no call count to quote.

Roadmap now **34/59** with AC-1, AC-2, AC-4, AC-7 and AC-11 closed. CI green.

PR #1765, opened 2026-08-31
roadmap: council-topology 18/77 to 26/77 — Phase 4 null-closed, Phase 2 costed at 20 UTC days (council 2/2)

matze4u:
Advances `road-to-inbox-harvest-2026-08-e-council-topology-evidence` from 18/77 to 26/77 done with 4 further steps null-closed, and produces the artefact the whole roadmap turns on: **the provider-call graph for Phase 2, which did not exist**.

## Steps closed (8) and null-closed (4)

| Step | Result | Evidence |
|---|---|---|
| **4.1** | `[x]` | Revisit record written — `agents/evidence/analysis/parallel-fanout-revisit-2026-08-31.md`. Verdict **keep sequential**; Phase 4 closes as a published null. A transcription of the already-resolved `blocker: parallel-fanout-reopens-a-closed-decision` (maintainer-owned, council 2026-08-28, 2/2), not a fresh decision. What changed since: **materially nothing** — the shipped example still enables two members, `on_overrun` is still wired at `orchestrator.ts:691-711`, and no latency complaint exists in the tree. |
| **4.2, 4.4, 4.5, 4.6** | `[-]` | NULL-CLOSED by 4.1. Executes a recorded maintainer-owned resolution (`Blocks: all of Phase 4`, resolution (a) *close Phase 4 as a published null*) rather than making a new one. |
| **4.3** | `[x]` | Verification-only, three readings: `orchestrator.test.ts` **83/83**, `git diff origin/main` on that file → **0 lines** (green *unmodified*, not green because edited into agreement), `grep -c 'Promise.all'` → **0**. |
| **7.1** | `[x]` | Topology vocabulary closed at seven names — `src/scripts/ai_council/topology_vocabulary.ts`. |
| **9.2, 9.3, 11.4, 12.4** | `[x]` | Verification/assertion closures, each with its grep or listing cited. |
| **2.1** | `[x]` | Twelve families pre-registered with per-family criteria and labels, before any arm runs. |
| **2.6** | `[x]` | `internal/bench/council-topology-promotion-stats-PREREG.md` — the reporting bar, written before any arm exists. |

### 7.1 — non-representability in the strong form

The type imports `LadderVerdict` (`judgment_ladder.ts:42-52`) and `ImpactClass` (`necessity.ts:545-550`) as **types only** — erased, zero runtime coupling — and asserts `[Extract<CouncilTopology, LadderVerdict>] extends [never]`, plus the same for `ImpactClass`. That excludes every ladder verdict and every impact class, not just the two named strings. A fourth assertion pins `team` and `user_required` back to those foreign unions, so an upstream rename reds this file rather than orphaning the exclusion into a dead string.

**Sabotage-proven, verbatim.** Appending an eighth member: `topology_vocabulary.ts(121,12): error TS2344: Type 'false' does not satisfy the constraint 'true'` (typecheck exit 2) and `arity: expected exactly 7 topologies, found 8` (vitest exit 1). Substituting `team` for `full_debate` — arity stays 7, so assertion 1 correctly stays green and only 2 and 5 fire: `reserved: 'team' is owned by another layer … and is not representable as a council topology`. `one_resolver_invariant.test.ts` **80/80 before and after**.

**Carrier decision, stated:** no JSON schema — `src/scripts/schemas/` has no council schema and `.ai-council.yml` is validated by a hand-rolled loader, so there is no enum surface to extend. Not `docs/contracts/ai-council-config.md` either: step 12.4 requires consumer surfaces to name capabilities, never topologies.

## Council decision — the Phase 2 benchmark (anthropic + openai, 2/2 convergent)

**Q1 = (a)** — a new `internal/bench/council-topology/` carrier. Not an extension of `adversarial-council/` (couples topology evidence to an adversarial corpus) nor of `council-blind-review/` (couples execution to a deferred human-rating workflow).

**Q2 = (iii) — Phase 2 as written is NOT executable**, because the roadmap defines experimental *dimensions* but never the provider-call graph. Neither `1,200 cells` nor `120 observations` is a valid provider-call total: host-solo may use zero external calls, current-council and full-debate are multi-call, ablation reuse is undeclared, "two axes" does not define how many model sets and topologies, and retries add calls. The ceiling is **50 calls per provider per UTC day**, not a fungible pool of 100.

**Both seats refused to weaken** — reducing the twelve families, deleting ablation, or treating unexecuted arms as nulls are **owner-reserved**. None was done.

**Q3** — zero-call work first, then 1B, then Phase 2 under a frozen manifest, then 3.3.

### What this branch built in response — all zero-call

`internal/bench/council-topology/` with `README.md` (carrier contract: what it owns, what it does not, the call graph with call-site citations, the overlap statement, the totals, the completion invariant, and the runner's preconditions recorded so building it cannot omit them), `PREREG-families.md`, and a generated `call-manifest.json` — **384 fully enumerated cells, 16 arms, 20 day-batches**. The frozen spec and its pure expander live in `src/scripts/ai_council/topology_bench_manifest.ts` with 24 tests.

**The number that did not exist:**

| | anthropic | openai | total |
|---|---|---|---|
| Minimum | 814 | 770 | **1,584** |
| Worst case (full retry reserve) | 924 | 880 | **1,804** |

**20 UTC days.** The arithmetic floor is `ceil(924/50) = 19`; day 20 is packing fragmentation, since a cell is atomic and never spans a day boundary. That is at **one item per family and N=2** — both declared, not derived. At three items per family it is roughly 60 days.

**The finding, reported and not acted on:** Phase 2 consumes both providers' entire daily quota for ~20 consecutive days, during which no other council work can run.

**A second finding:** at N=2 this benchmark licenses **no promotion claim at all**. The already-committed 2.6 floors are n≥5 / n≥10; N=2 clears neither, so Phase 2 produces descriptive comparison only. That is 2.6 read honestly.

**Overlap, stated explicitly:** exactly two arms reuse `baseline-default-council`. Everything else re-runs — the ablation ladder deliberately does not borrow a baseline's generation output, because a rung borrowed across a UTC day carries day-as-confounder into the one arm whose purpose is attribution.

**"Not run is not a null" is encoded, not asserted** — sabotage-proven in both layers. Adding `'pending'` to `PHASE2_COMPLETE_STATUSES` → `topology_bench_manifest.ts(100,36): error TS2344` plus a module-load throw that stopped the test file collecting. `BENCH_FAMILY_ARITY = 11` → `error TS2367` and `expected exactly 11 pre-registered families, found 12`.

### The twelve families — all quoted, none invented

Step 2.1 enumerates exactly twelve in its own prose; every one is a quotation in the roadmap's order. The **label assignment** is this run's classification, with the discriminator stated per family: a deterministic key or executable oracle means confirmatory, a rubric-judged component means exploratory.

- `gradeable-confirmatory` (6) · `model-graded-exploratory` (5) · `human-rubric-deferred` (1)

The deferred family is registered in full and runs no arm — 32 cells marked `not_eligible` rather than dropped, with **no model-graded substitute in its place**, per the resolved `blocker: maintainer-blind-ratings` option (b).

## Honest limits, stated by the work that produced them

- **The call graph is a frozen contract, not a measurement.** Every pass is cited to a call site (`orchestrator.ts:394`, `:1553`; `consensus_round.ts:107`, `:141`, `:191`), but no council was run to count actual calls. If the runner deviates, the manifest is right and the runner is wrong — the intended direction, but an assertion about a runner that does not exist.
- **"Current default council path" rounds are pinned per arm, not read from config** — the runtime defaults disagree with each other (`consult` rounds=1, `run_debate` max_rounds=2, the live run used 3), so that arm models a configuration this run chose rather than one it observed.
- **`MS2-cheap` is asserted to exist**; `.ai-council.yml` was not read to confirm a cheaper pinned tier is configured for both providers. Strongest/cheapest provider assignment is a pin, not a verified ranking.
- **Ablation rung semantics are a declared design**, not a mapping onto existing code paths — no `judge_synthesis` or `peer_review` executor exists to check against.
- **12.4 is vacuous** — no topology vocabulary is implemented, so no command file *could* hardcode one. Closed as a baseline taken before the vocabulary exists, and the note says so. One near-hit recorded so a later grep does not re-litigate it: the `judge-synthesis` **skill** is a report-shape name predating this vocabulary.
- **11.4's pin is a naming-based shape gate over source text**, not a module-graph proof; a loader named outside the pattern set escapes it. It tests the denial as well as the claim (4 violating snippets must match, 3 ordinary lines must not).
- **9.2 closes as "no violation present"**, not as a guarantee — the fan-out lane 9.1 describes does not exist.
- **7.1's compile-time layer has no automated test that runs `tsc`.** Its enforcement is `npm run typecheck`, confirmed to run in CI at `.github/workflows/tests.yml:285` and proven red under both sabotages. The test file adds a structural guard so the assertions cannot be silently deleted; a vitest run never type-checks and this is not pretended otherwise.

**One self-correction worth naming.** Step 4.3's evidence originally cited `git diff origin/main --stat -- tests/` → empty. That went false the moment the 11.4 pin test was added. Re-scoped to the orchestrator file (still 0 lines) with the added file named — commit `8ffa6584e`.

## Not attempted, with the reason

1B.1 and 1B.4 need ~40-60 provider calls and this branch spent **zero**. 3.3 stays blocked on the open `leakage-bench-needs-quota-and-an-uncommittable-corpus` — its second half survives the quota reset, since `agents/runtime/council/` is gitignored and auto-pruned so the ≥30-body corpus must be assembled locally at measurement time. 2.2-2.5 and 2.7 are downstream of the manifest, which is a precondition and not their execution. Phases 5, 6, 8, 10-13 need decisions this run did not put to a council.

## Gates

`lint_roadmap_blockers` · `lint_roadmap_complexity` · `lint_roadmap_ci_steps` · `check_roadmap_trackable` · `check_no_roadmap_refs` · `lint_empty_roadmaps` · `lint_roadmap_later_disposition` · `check_estate_count` (+0 active / -0 disposed) · `check_source_size_budget` · `lint_evidence_artifacts` · `lint_plan_risk_register` · `check_references` · `check_council_references` · `check_condensation` · `check_no_external_sources` · `check_md_language` · `npm run typecheck` · `eslint` — **all exit 0**. `tests/scripts/ai_council/` suite: **1,219 passed**. The manifest re-emits byte-identically.

`task ci` was not run locally — the repo's own notes record it red on `main` without a branch-vs-base diff. Remote CI is the authoritative gate.


matze4u:
## Update — 1B.1 was run live and did NOT close, plus the run-8 summary

Two more commits since the PR body.

### Step 1B.1 — run, half met, **left open**

The council of 2026-08-31 ranked 1B first among the roadmap's three quota consumers, and the quota was available, so it was run. It stays `[ ]`, and that is the correct outcome rather than a shortfall.

**Call accounting, predicted before spending and then measured:**

| | predicted | actual |
|---|---|---|
| anthropic | +2 | **+2** (24 → 26) |
| openai | +2 | **+3** (24 → 27) |
| total | 4 | **5** |

The extra openai call *is* the finding.

**Half 1 — the block parses inline: MET, on the anthropic seat.** Its persisted `raw_text` ends in a fenced JSON array of five `{"id","text"}` objects, and its `text` ends in the `_[inline findings block extracted: 5 item(s); …]_` marker — which `harvest_inline_findings` writes **only after** both `parse_findings_outcome` and `_isOwnFindingsBlock` pass. That is a parse receipt, not a fence sighting. The five consensus findings under that member carry the same five ids in the same order.

**Half 2 — zero extraction calls: NOT MET at the run level.** The openai (`codex-default`) seat emitted **no block at all** — a prose "Top-5 consensus" bullet section, no fenced block, no bracket array anywhere — so the 1B.2 repair path fired. `consensus.extraction_responses` holds exactly one entry, `provider=openai`, and that seat's counter delta is 3. Proven from the recorded response set **and** the counter, not by reading the code and concluding it should be zero.

The step's own condition says zero. One is not zero, so the checkbox does not move.

**What the run does settle**, recorded so it need not be re-established: the config chain that the 2026-08-30 defect broke resolves live (probed **for free before spending** — `inline_findings: true` survives file → `synthesizeAiCouncilBlock` → `inlineFindingsActive`, and reads `false` for every non-analysis lens); the contract reaches the final round; the locator finds a real model's block; strip and marker fire; harvested findings reach consensus with no second call for that member. What is missing is a run where *every answering seat* carries the block.

**Recorded as an observation at n=1, explicitly NOT as a 1B.4 datum** — no rate, no matched comparator, arms not started. Whether the openai miss is systematic is unknown and the artefact says so.

One setup note worth carrying forward: the worktree had no `agents/runtime/state/council-probes.json`, which would have produced a degraded 0/2 run **and still spent quota**. It was copied from the main checkout before spending.

### The drain-run summary

`agents/evidence/drain-run-summary.md` is rewritten as the run-8 record (git history keeps run 7): three PRs, six council rounds, two transfers, one null-close, and **twelve defects the execution found that the roadmaps did not** — including a frozen holdout `SET-SHA256` that does not recompute over a byte-identical file set, an acceptance criterion falsified months ago by a different roadmap, a resume probe that reports `fired` on an archived roadmap whose criterion was transferred unresolved, and this Phase 2 benchmark pricing out at 20 UTC days of exclusive quota.

It also records, plainly, that the roadmap directory is **not** empty and why: a council verdict required splitting owner-blocked work into its own ACTIVE receiver rather than parking or dropping it, so the run ends with four active roadmaps where it found three.

**Council quota spent by the whole run: 26 anthropic, 27 openai**, against 50 per provider per UTC day.

CI green.

PR #1819, opened 2026-09-03
roadmap: complete road-to-artifact-location-and-doctor-reach

matze4u:
## What this is

`road-to-artifact-location-and-doctor-reach` executed to completion and archived.
A roadmap-shaped file written outside `agents/roadmaps/` is now a finding, and
`agent-config doctor` can return a non-zero exit.

Verified against `2b3d2b347`. 9/9 steps, 5/5 acceptance criteria.

## Council decision standing in for maintainer sign-off

`doctor-exit-contract` was an owner-reserved Class-3 A/B choice. Under the drain
mandate it went to the AI council: 2026-09-03, members
`anthropic/claude-sonnet-4-5` and `openai/codex-default`, three rounds, blind
chairman. **Unanimous (b)** — `doctor` keeps its exit contract and `--strict`
carries the failing exit — agreeing with the roadmap's own recommendation. Both
seats: ADR-041 governs verb growth so a flag beats a verb, and the existing
contract is load-bearing for callers that run `doctor` for information rather
than for permission.

Both attached the same two conditions, adopted in full:

- **A separately configurable severity threshold** → `--strict-level fail|warn`,
  default `fail`.
- **An immediate way to disable failing exits that is not a code rollback** →
  `AGENT_CONFIG_DOCTOR_NO_FAIL=1`. It suppresses the *exit code*, never the
  finding, and announces the suppression on stderr. A switch that hid the
  finding too would be the ignored-warning shape this repository already
  documents.

## Phase 1 — the location gate

`src/scripts/check_agent_artifact_location.ts`. `is_roadmap_candidate` is
**imported** from `update_roadmap_progress` — the way `check_estate_count.ts:147`
does it, not mirrored the way `check_roadmap_trackable.ts:110` does — so the
roadmap corpus and this guard cannot drift apart.

**The name test alone is not the detector, and that is the substantive finding.**
`is_roadmap_candidate` answers "is this filename eligible *inside* the roadmap
root": it excludes four names and one prefix and otherwise accepts nearly every
`.md`. Run bare over this repository it would report thousands of legitimate
documents — Risk 1 of the roadmap's own register. So the name test is a
pre-filter and the decision is `roadmapShape`: roadmap frontmatter **and** a
`## Phase` heading **and** a checkbox step. All three, because each alone is
common — plenty of docs carry frontmatter, and a design doc can carry checkboxes.

The finding names which signals fired, so a false positive is arguable rather
than mysterious.

**Polarity proven in both directions**, which step 1.1 required in as many words.
The `--self-test` runs 5 cases (2 rejecting) against the real CLI: a
roadmap-shaped file under `docs/roadmaps/` and one at the tree root must red; the
byte-identical file under `agents/roadmaps/` and under `agents/roadmaps/later/`
must stay green, and so must prose carrying frontmatter and a checkbox but no
phase heading.

**Sensitivity probe:** loosening the predicate from three-of-three to
one-of-three turns **6 of the 14** vitest cases red; restoring it returns 14
green.

**Registered** in `src/config/gate-coverage.yml` with `min_scanned: 3000` against
a real 3,866, and a corpus note recording that the unit is every `.md` the walk
*visits*, not the ones it matches — zero matches is the normal clean state, so
only the unfiltered walk separates "nothing misplaced" from "nothing read". The
manifest was **recounted on this tree** rather than incremented:
`count_gate_scripts()` returns **313**, row count **84**, so 26.8 %. The script
count had risen by three since the last recount without one.

**The canary reaches this gate exactly** — planting a roadmap-shaped file outside
the root *is* the violation, so unlike several neighbours it needs no
`no_canary_reason`. Verified: `check_gate_coverage --canary` reports the plant
caught at exit 1, and the tree is clean afterwards.

**1.3 — the reach limit is in the gate's own header**, at length: this gate
protects this repository, the observed failure was in a consumer repository, and
a consumer never runs this package's CI, so a gate scoped here protects the tree
where the mistake is *least* likely.

## Phase 2 — `doctor`

**Finding: the roadmap's premise was half wrong, and the correction is the work.**
The Context says `doctor` "writes nothing and always exits zero", citing two
independent reviews. Measured at `2b3d2b347`: the default run already returns 1 on
manifest **drift** (`return drift_present ? 1 : 0`), and `--ci` already folds
check failures in on top. What genuinely did not exist is a failing exit for check
failures **outside the `--ci` JSON contract** — `--ci` sets `opts.json = true`, so
a human-readable strict run was impossible. That gap is what `--strict` closes,
and it is a narrower change than the roadmap assumed.

**Second finding, from the implementation: `main()` has two exit paths and only
one is obvious.** A repository with no install manifest returns through
`_run_no_manifest` long before the drift branch. The first `--strict`
implementation lived only in the drift branch and exited **0** on this very tree —
which carries a `fail` check row and emits `can_proceed: false`. The verdict is
factored into `_strictExit`, called from both paths. This is the same class as
the postmortem the roadmap came from: a mechanism keyed on a path the artefact did
not take.

**2.2** — `--json` already existed; `can_proceed` did not (`grep -rc can_proceed
src` returned zero hits tree-wide, exactly as the Context recorded). The payload
now carries `can_proceed` (boolean) and `status` (`ok`|`warn`|`fail`). Both are
facts about the tree and deliberately independent of `--strict`: a caller running
with the kill switch set still reads `can_proceed: false`, because the exit code
is a policy and `can_proceed` is not.

**2.3 — the caller enumeration, with a disposition each. Every one is
"unaffected", which is the point of option (b).**

| Caller | Kind | Disposition |
|---|---|---|
| `src/scripts/_cli/cmd_conformance.ts:57` | imports the check runner, runs its own `--ci` semantics | unaffected — `--ci` untouched |
| `src/ui/wizard/FinishChecklist.tsx:25` | displays the command for the user to run | unaffected — no exit-code consumption |
| `src/scripts/install.ts:508`, `:4378`, `:5044` | remedy hints | unaffected — prose |
| `src/scripts/install:501`, `src/scripts/sync_agent_settings.ts:78` | remedy hints | unaffected — prose |
| `src/domains/meta/team/**/command.md` (6 sites) | instruct `doctor --check team` | unaffected — `--check` path untouched |
| `src/scripts/condense.ts:2185`, `src/config/surface-matrix.yml:5` | generated / config prose | unaffected |
| `.github/workflows/**` | **no workflow invokes the verb** — the two matches are comments naming `reach:doctor` and a crash class | unaffected |

Zero callers consume the plain exit code today, and none opts into `--strict`
here. Adoption is a later per-caller decision, which is what "leaving current
callers untouched" means.

Behaviour verified end to end: plain `0` (unchanged), `--strict` `1`, kill switch
`0` with the suppression announced, unknown `--strict-level` `2`.

**The CLI test cases derive their expectation from the `--json` payload** rather
than hardcoding today's tree state, so repairing the currently-failing check does
not red a correct test.

## Phase 3 — the discriminator that was missing

Two invariants added to `docs/contracts/agents-layout.md`:

**The reader decides the directory** — `docs/` is product documentation for
humans, `agents/` is agent working material. And the sentence that matters: **a
missing `agents/` in a target repository is a directory to create, never
counter-evidence for a `docs/` convention.** That inference is what produced the
incident; the absence of the directory is the state before the first artefact,
not a signal about where artefacts go.

**An artefact has no authority over its own location** — a location note inside a
file is a prior session's intent, to be checked against the contract. Where they
diverge, the divergence is the finding.

The second is marked **`instruction-only`**, deliberately and explicitly. No
mechanism reads it, and the postmortem is right that it would not have stopped the
run it describes — that agent read a location note, acted on it, then defended the
result. It is carried because it generalises past location, and because a rule
nobody wrote down cannot be cited by the next reviewer either.

**3.3 — not built.** The postmortem proposes a cross-repo artefact-move trigger.
One observed instance does not justify a new command surface.

## Gates run

| Gate | Result |
|---|---|
| `check_agent_artifact_location` | clean, 3,882 markdown files read |
| `check_agent_artifact_location --self-test` | 5/5, 2 rejecting |
| `check_gate_coverage` | every enforced gate cleared its floor |
| `check_gate_coverage --canary` | plant caught at exit 1, tree clean after |
| `lint_roadmap_blockers` | 0 violations |
| `lint_roadmap_complexity` | **red, pre-existing** — 5 siblings, this roadmap green |
| `check_roadmap_trackable` | 0 violations |
| `check_no_roadmap_refs` | clean |
| `lint_empty_roadmaps` | clean |
| `lint_roadmap_later_disposition` | clean |
| `lint_carrier_integrity` | scanned 660, planned 660 |
| `check_estate_count` | within ratchet |
| `tsc -p tsconfig.json --noEmit` | clean |
| `tsc -p tsconfig.scripts.json` | clean |
| `vitest` (location gate + doctor strict + doctor) | 134/134 |
| `task sync` + `task generate-tools` | no projection drift |

`lint_roadmap_complexity` is red on clean `origin/main` for the lightweight
roadmaps whose `relates:` rows use the old slug-only form. This branch fixes its
own; the rest belong to the runs that execute them.

## Honest nulls

- **No cross-repo move trigger was built.** 3.3 asked for that, and the absence is
  the deliverable.
- **The artefact-authority clause is enforced by nothing**, and says so in the
  contract text rather than being implied to be enforced.
- **The location gate protects this repository only.** A consumer never runs this
  CI. `doctor --strict` is the surface that can carry the finding outward, and the
  gate header states the limit so it cannot be cited as broader than it is.


matze4u:
## Triage of the one blocking security finding

### `412040920bb4` (critical) — FALSE POSITIVE

The claim: the CI self-review runs against a prompt that instructs subset
reporting, creating an invisible withholding risk. It is refuted on three
independent grounds, any one of which is sufficient.

**1. The cited file is not in this branch's diff.**

```
$ git diff --name-only origin/main...HEAD | grep adversarial-review
$ echo $?
1
$ git diff --stat origin/main HEAD -- src/skills/adversarial-review/SKILL.md
          # empty — byte-identical to origin/main
```

13 files change on this branch and none of them is under `src/skills/`. This
branch introduces no prompt text at all.

**2. The file's current text forbids subset reporting — twice.**

`src/skills/adversarial-review/SKILL.md:66` — *"**Debate internally, report
completely.** … every criticism raised reaches the report with its disposition
(must-fix / deferred / rejected, each with one line of reason). Do not decide on
the user's behalf which concerns were worth their attention."*

`:161` — *"**Every criticism raised**, each with severity and disposition …
Not "top concerns" — a summary that keeps only the highlights is the pre-filter
defect wearing an editor's hat."*

The word `subset` occurs exactly once in the file, at `:74`, inside the sentence
that documents the **removed** wording: *"This clause used to read 'surface only
the trade-offs that need the user's input', and that is the pre-filter defect: an
instruction to report a subset is followed literally."* That passage then names
this very mechanism as the reason it matters — *"`self_review_gate.ts` loads this
file as the system prompt for the package's own CI self-review, so a suppression
clause here suppresses findings on every pull request."* The finding restates the
file's own argument back at it as if the fix had not been applied. It was applied
on 2026-08-06 in `c4a49204d` and is on `main`.

**3. The consuming gate applies no cap either.**

`src/scripts/self_review_gate.ts:parseFindings` keeps every finding whose
severity and kind are in the valid enums — it truncates the `title` and `detail`
*strings* (300 / 2000 chars) and drops nothing. `renderReview` then maps over
`findings` in full into both the rendered table and the
`release-findings-json` machine-readable block. The only `.slice(0, 12)` in the
file is `findingId`'s sha256 hex truncation at `:66`, not a findings limit. The
three comments this gate posted on this PR carried 6, and on its siblings 8, 12
and 10 findings, with no top-N anywhere.

No fix, because there is nothing to fix: the withholding instruction the finding
describes does not exist in the prompt, in the gate, or in this diff.

The remaining findings on this PR are all marked Advisory and are outside the
scope of this pass, which covered the blocking-severity security set.

PR #1821, opened 2026-09-03
roadmap: complete road-to-cascading-base-integration

matze4u:
Completes `road-to-cascading-base-integration` (14/14 steps, AC-1…AC-5), resolves its `cascade-default-inclusion-policy` blocker, and archives it.

## The council decision, verbatim

The blocker asked whether the default branch belongs in `resolveBase`'s integration set when the PR target is a release line. It offered two answers: **(a)** always include, **(b)** include past a commit-distance threshold. It went to the AI council on **2026-09-03**, members `anthropic/claude-sonnet-4-5` and `openai/codex-default`, standing in for maintainer sign-off under the maintainer's standing drain mandate.

**The answer is neither (a) nor (b). It is a third option: an explicit per-target branch-convergence policy, read from the resolved PR target commit SHA.**

**Round 1 split on the offered fallbacks, and it was not a split on substance.** Both seats independently named the *same* third option as correct, and picked opposite fallbacks only because the framing demanded a fallback.

- Seat 1: *"Reviewer B's policy-based third option is strongest."*
- Seat 2: *"The correct third option is an explicit per-target branch-convergence policy based on repository intent, not commit distance."*

Both rejected (b)'s commit-distance threshold as an unmeasured constant — the same objection the blocker's own `Recommendation:` field already carried.

**Round 2 was a focused follow-up that specified it, and the two seats converged on a near-identical design.** That specification is the decision, and it is what this PR builds:

1. **Where the policy lives.** Section `branchConvergence`, loaded **exclusively** from the resolved PR target commit SHA. The caller resolves and supplies `targetSha`; the loader never infers it from the working tree or a branch name. Both seats called this non-negotiable — reading policy from the PR head, the checkout, or an unpinned branch ref lets a reviewed change edit its own acceptance criteria, the canonical confused-deputy bypass. One seat: *"I would block any implementation reading policy from PR head, checkout state, or unpinned branch name."*
2. **Shape.** `enabled: true | false`; `targets` keyed by **exact** branch name; `defaultBranch: include | exclude`. Unknown fields, invalid values and duplicate keys rejected. Globs, pattern precedence, inheritance, configurable reason strings and any repository-wide fallback were cut explicitly for v1 by both seats.
3. **Missing entry fails.** `MissingBranchConvergencePolicy(target="release/1.x")`, never a partial set. Neither inclusion nor exclusion is universally safe, and an absent entry must not manufacture repository intent. A default-branch target needs no entry.
4. **One stable result type.** `{ entries: Array<{ref, reason}>, policyStatus }`, `reason` a fixed **closed** set, never user-written prose — both seats rejected free-form reasons as fixture churn plus a log-injection surface. Returning a bare array normally and an object when disabled was refused as coupling every consumer to ad-hoc type discrimination.
5. **Kill switch.** `enabled: false` → `policyStatus: "disabled"`, target only, surfaced as **BYPASSED, never as passed** — a stderr warning alone is insufficient because callers discard stderr. *Activation criteria:* policy evaluation causes materially incorrect blocking; target-SHA config cannot be loaded reliably; an incident requires restoring pre-policy behaviour. *Removal:* re-enable once the defect is fixed and the failing fixture passes.
6. **Twelve fixtures**, the union of both seats' lists. 7, 8 and 9 non-negotiable.

Both seats confirmed the work fits `complexity: lightweight` with the v1 cuts, and both stressed that step 1.1 *is* the contract Phases 2–4 consume, so its result and error shape had to stabilise first. It did — Phases 2–4 were built against it unchanged.

## Two deviations, recorded rather than smoothed over

**1. The policy file is not `.agent-settings.yml`.** The council named it as "the existing committed" file. It is not committed anywhere this package runs: `.gitignore:317` ignores it in this repo, and `src/config/gitignore-block.txt:56` writes that same ignore into **every consumer's** managed block. So `git show <sha>:.agent-settings.yml` returns nothing at every SHA in every repository this ships to, and the feature would have been dead on arrival. A SHA-pinned read requires a **tracked** file, so the path moved to `.branch-convergence.yml` and nothing else moved: same section name, same shape, same exclusive read from the target SHA. The architecture the council called non-negotiable is unchanged; only a premise about where it could live turned out to be false. `src/scripts/_lib/branch_convergence.ts:26-42` carries the evidence.

**2. Entry order vs integration order.** Council § 4 and fixture 4 pin the **result** as target-first; roadmap step 1.3 pins **integration** as default-first. Both are honoured because they order different things: `entries` is target-first (the target is the ref the PR merges into, and the stable head of the contract), and `integrationOrder()` merges the default first so the broad conflict surfaces where it is cheapest to abandon. Neither is inferred from the other; both are asserted, at `sync_pr_branch.ts` `integrationOrder` and in the `1.3` fixture.

## Steps, with provenance at this commit

| Step | Evidence |
|---|---|
| 1.1 set from `resolveBase` | `src/scripts/sync_pr_branch.ts` `resolveBase` → `ResolveBaseResult`; policy in `src/scripts/_lib/branch_convergence.ts` |
| 1.2 reason per entry | `BaseReason` closed set + `describeReason`; fixture *1.2 — every entry carries its own resolution reason* |
| 1.3 stated integration order | `integrationOrder` (default first) + `renderBaseSummary`; fixture *1.3 — integration runs the DEFAULT first* |
| 2.1 one invocable target | `taskfiles/dev.yml` `push-ready:` — six steps, `DRY=1` read-only. Dry run executed on this branch; output in § Dry run below |
| 2.2 hook refuses, never merges | `src/scripts/install-hooks.sh` — `THIS HOOK REFUSES. IT NEVER MERGES, AND NO ONE MAY "IMPROVE" IT INTO ONE.`, pointing at `task push-ready`. Test asserts no `git merge/rebase/pull` line was added (`merge-base` is a read-only query and the only permitted occurrence) |
| 3.1 auto-resolve GENERATED | `autoResolveGenerated` — regenerate, stage, **regenerate again**, refuse if any path is still dirty. Byte-identity is asserted, not assumed |
| 3.2 never REMEASURED/AUTHORED | one non-generated path refuses the **whole** set; the test asserts the regenerator is called **zero** times on a mixed set |
| 4.1 pin base OID, retry ≤ 3 | `integrateWithPinnedBase` + `MAX_BASE_ATTEMPTS = 3`; every attempt's before/after OID is in the message |
| 4.2 no merge-train semantics | asserted by equality of two identical calls — a queue or `rerere` cache would make the second differ |

**Anchors verified at `2b3d2b347`, all four exact:** `check_branch_freshness.ts:274` `resolveBase`, `:160` `askForgeForBase`, `sync_pr_branch.ts:238` `classifyConflicts`, `sync_pr_branch.ts:10` (PR #1391). **One drifted:** the roadmap's Context table writes the hook as `install-hooks.sh:24-155`; the file is `src/scripts/install-hooks.sh` and is 373 lines, not ~155. Path and range both stale; the cited content (the `cat > pre-push` heredoc opening at `:24`) is where the roadmap says it is.

## The twelve fixtures

`tests/scripts/branch_convergence.test.ts`. **All twelve were observed RED before green** — the first run reported `12 failed | 1 passed`; the one green was F10, whose pure-parse half was already satisfied because the loader module landed before the test file, so F10 got its own sensitivity probe instead.

| # | Fixture | First observed |
|---|---|---|
| 1 | default target → default ref only, `not-required` | RED |
| 2 | missing non-default entry → typed error, no partial set | RED |
| 3 | `exclude` → target only | RED |
| 4 | `include` → target then default, exact reason codes | RED |
| 5 | kill switch → target only, `disabled` | RED |
| 6 | gate renders `disabled` as BYPASSED, not passed | RED |
| 7 | **trust boundary** — a policy in the PR head does not override the target-SHA policy | RED |
| 8 | **negative** — the `include` fixture fails if the default ref is omitted | RED |
| 9 | **negative** — the `exclude` fixture fails if legacy logic adds the default unconditionally | RED |
| 10 | invalid / unknown policy values fail schema validation | green on first run (module preceded the test); probed below |
| 11 | unresolvable target SHA fails closed | RED |
| 12 | target/default identity produces no duplicate entries | RED |

### Sensitivity probes

Each probe neutralised the mechanism, was observed red, and the file was restored **byte-exact** (`cmp` verified).

| Probe | Neutralisation | Result |
|---|---|---|
| **F8** | deleted the `include` push of the default entry | F4/F8 red |
| **F9** | made the `exclude` return conditional so the default is appended anyway | F3/F9 red |
| **F7** | read the policy at a SHA other than the target's | F7 red — both the `exclude`-wins assertion and the *reader was asked for exactly `[TARGET_SHA]`* assertion |
| **F10** | removed the unknown-field rejection | F10 red |
| symref regression | forced `parseSymrefDefault` to return `null` | both its cases red |
| 2.1 / 2.2 | renamed step 5 in the task; softened the hook sentence | both Phase-2 cases red |

## A defect the step-2.1 dry run surfaced

Running `task push-ready DRY=1` on this branch printed step 2 refusing with `no open PR and no origin/HEAD` while step 5, three lines later in the *same documented sequence*, passed with `branch is current with origin/main`. Cause: `sync_pr_branch` read the default branch from `refs/remotes/origin/HEAD` alone — a clone-time ref that is simply unset in this checkout — while `check_branch_freshness.ts:223` `serverDefaultBase` asks the server. Two resolvers disagreeing about the base is this roadmap's own subject one layer down, and the default branch stopped being cosmetic the moment the policy's `include` needed a ref to add. Fixed by mirroring the server symref read; extracted as the pure `parseSymrefDefault` so the parse is covered without a network round trip.

## Dry run (step 2.1 verify)

```
1 fetch                     → origin fetched
2 integrate the base set    ✅ already current with every base ref — no branch-convergence
                               policy required (target is the default branch);
                               integrating in order: origin/main (the repo default branch).
3 regenerate                → would run: task sync && task generate-tools
4 verify                    → would run: task consistency
5 re-check freshness        ✅ branch is current with origin/main (2b3d2b347)
6 push                      → NOT run here. Review the merge, then push deliberately.
```

## Honest nulls

- **This repository has no release line and no non-default PR target.** `git branch -r` → 405 branches, zero `release/*`; `git symbolic-ref refs/remotes/origin/HEAD` is unset. So the *only* live path here is `policyStatus: "not-required"`, and every policy fixture is necessarily a unit fixture over injected git answers. A live fixture could only ever have covered the default-target case.
- **No `.branch-convergence.yml` is added.** With no non-default target, an empty policy file would change nothing: absent-at-the-target-SHA and present-but-no-entry produce the identical `MissingBranchConvergencePolicy`. Adding a file nothing reads would be an artefact for its own sake. The path is named and documented in the loader; a maintainer opening a release line writes it then.
- **`check_branch_freshness.ts`'s own `resolveBase` is untouched.** Gap A names `sync_pr_branch.ts:292-318` precisely, step 1.2 names its `how` field, and Phase 2 integrates — which is `sync_pr_branch`'s job; `check_branch_freshness` only detects. The Context table cites the other `resolveBase` as *present* capability, not as the thing to change.
- **Step 6 of `push-ready` does not push.** The documented sequence ends in a push; the runner prints what to do and stops. Pushing is a Hard-Floor action and a task target that ends in `git push` would make it a side effect of running a diagnostic.
- **A completion-review artifact is not included.** `check_completion_review` reports one advisory `missing-artifact` finding on push (5 code paths of 11 changed files). Advisory, not blocking; stated rather than skipped with a declaration.

## AC-5 verified

`no rerere configuration, no queue state, and no new observability layer exists as a result of this roadmap.` — `grep -rn rerere src/ scripts/ taskfiles/ tests/` returns only two prose lines that *say there is none*; `grep -n 'writeFileSync|mkdirSync|appendFile'` over both new/changed modules returns nothing. Phase 4's bound is a retry with a ceiling held entirely in the run.

## Gates

| Gate | Result |
|---|---|
| `lint_roadmap_blockers` | PASS (+ decidability 0 violations) |
| `check_roadmap_trackable` | PASS |
| `check_no_roadmap_refs` | PASS |
| `lint_empty_roadmaps` | PASS |
| `lint_roadmap_later_disposition` | PASS |
| `lint_carrier_integrity` | PASS |
| `check_estate_count` | PASS |
| `lint_code_comments` · `lint_output_slop` · `lint_originality` · `check_source_size_budget` | PASS (946 lines, under the 1500 cap) |
| `npx tsc -p tsconfig.json --noEmit` · `npx tsc -p tsconfig.scripts.json` | PASS |
| `task consistency` | PASS |
| vitest — `branch_convergence` (28) · `sync_pr_branch` (24) · `check_branch_freshness` (18) · `prepush_delete_only` · `prepack_lifecycle_check` · `preflight_single_delivery_binding` | 88 passed |
| `lint_roadmap_complexity` | FAIL on **5 sibling roadmaps**, pre-existing — 6 failed before this branch, 5 after, because this branch fixed only its own `relates:` rows. Their rows are byte-identical to `main`. |

## `relates:` frontmatter (task item 3)

Rewritten to rule 18's `slug:` + `relation:` + `note:` schema, bare slugs, each classified from the target's own text — never invented, and no `depends` (so no `depends:` mirror is owed):

- `road-to-merge-hotspot-drawdown` — **extends**. It built the GENERATED/REMEASURED/AUTHORED split at `sync_pr_branch.ts:238`; Phase 3 extends exactly that classification into auto-resolution. Archived and complete, so no execution dependency.
- `road-to-merge-surface-zero` — **extends**. Shares `src/scripts/sync_pr_branch.ts` and its test file, and its own (closed) 1.1 extended the same GENERATED list. Declared rather than called disjoint because the file overlap is real; its merge-queue and observability half is explicitly *not* extended here — AC-5 forbids both. Parked in `later/`.
- `road-to-merge-confirmation-doctrine` — **disjoint**. Shares the word "merge" and nothing else: it is about whether a direct `merge PR #123` is itself the Hard-Floor confirmation, a kernel-rule authority question. Verified by grep — it names none of `sync_pr_branch`, `check_branch_freshness`, `install-hooks` or `resolveBase`.

## Post-open CI fixes

Two real failures on the first CI run, both fixed on this branch and verified locally before the re-push.

**1. `Rule backstops` — `check_gate_reachability`.** `push-ready` is gate-shaped by the checker's command predicate, because step 5 of the sequence calls `check_branch_freshness`. Wiring was the wrong answer: the target **mutates** — it merges the base set in and regenerates derived files — so wiring it would make CI perform merges, which is precisely what step 2.2 forbids the pre-push hook from doing one layer down. Recorded as an exemption row in `src/config/gate-reachability-exemptions.json` with the reason and the honest promotion condition (*nothing* would make it a gate; the checks inside it are wired to CI separately and on their own). `check_gate_reachability --gate` → 15 exempt, green.

**2. `Sync + Generate Tools Consistency` — `check_portability`.** One violation: `` `task push-ready` `` in `dist/agent-src/commands/pr/create.md:152`. That file is **projected to consumers**, and a consumer project has no Taskfile and no `push-ready` target — the rule caught a real defect, not a false positive. The shipped paragraph now describes the six steps and the dry mode, names where *this package* wires them, and says what a consumer substitutes for the two generator steps. The concrete runner name stays in the two maintainer-side surfaces that are not projected, `taskfiles/dev.yml` and `src/scripts/install-hooks.sh`. `check_portability` → 1694 scanned, clean.

**3. `Static Checks` — not this branch's.** `npm audit --omit=dev --audit-level=high` reports three pre-existing advisories (fast-uri high; hono/qs moderate) that are red on `main` and on every branch cut from it. `git diff origin/main -- package-lock.json package.json` is **empty** on this branch; PR #1822 carries the lockfile bump, and duplicating it here would only conflict on merge.

## Live evidence for AC-1 and 4.1

While integrating `main` before the re-push, `sync_pr_branch` was used on itself rather than `git merge`:

```
✅  no branch-convergence policy required (target is the default branch);
    integrating in order: origin/main (the open PR base).
    integrated origin/main on attempt 1.
    REGENERATE derived files now — a clean auto-merge of a generated file is still wrong.
```

It resolved the base from the open PR, classified the target as the default branch (`policyStatus: "not-required"`, one entry), pinned the base OID and integrated on the first attempt. The `not-required` path and the Phase-4 pinning are therefore exercised against a real remote, not only against fixtures.

## Chunking note

The archived roadmap file itself landed in `8315ee9c1` rather than the roadmap commit, because the rename was already half-staged when that commit was made. Recorded rather than rewritten: `git-history-discipline` puts commit-chunk order below not rewriting history, and the final commit carries the other half.


matze4u:
## Triage of the two blocking security findings

Guard added in `4a6544c96`. No behaviour change on either path.

### `b64b04412839` (critical) — FALSE POSITIVE as framed; one real premise underneath it, now pinned

The framing is inverted. The deviation is what **keeps** the trust boundary,
not what breaks it. The council specified `.agent-settings.yml` § `branchConvergence`
on the premise that it is committed. It is not — `.gitignore:317` ignores it
here, and `src/config/gitignore-block.txt:56` writes that same ignore into every
consumer's managed block — so `git show <sha>:.agent-settings.yml` returns
nothing at every SHA in every repository this ships to. The only way to read a
gitignored file is the working tree, which is precisely the confused-deputy
bypass one seat said it *"would block any implementation"* for. Moving to a
tracked path is what makes the SHA-pinned read possible at all.

The boundary itself holds, verified in the tree rather than from the docblock:

- `src/scripts/sync_pr_branch.ts:483` — `loadPolicyAtSha(targetSha, deps.readAtSha)`, where `targetSha` comes from `deps.remoteSha(target.ref)`, i.e. `git ls-remote --heads origin <bare>` — the server's answer, not the checkout's.
- `makeGitDeps.readAtSha` — `git show <sha>:<rel>` and nothing else; `branch_convergence.ts` imports only `yaml` and has no filesystem or child-process access at all.
- `ShaFileReader = (sha, repoPath) => string | null` offers no argument for the working tree, so a caller cannot reach the PR head by accident.
- Absence fails **closed**: `resolveBase` throws `MissingBranchConvergencePolicy` rather than defaulting, and a PR targeting the default branch needs no entry (identity by ref name or by SHA). *"An absent entry must not manufacture repository intent."*

What *was* genuinely unguarded is the deviation's own premise: nothing asserted
that `.branch-convergence.yml` is trackable. It is not ignored today
(`git check-ignore` exits 1 here, and in a scratch repo seeded with the managed
block), but an ignore pattern reaching it later would make the policy unreadable
at every SHA — and the tempting repair for *that* is a working-tree fallback,
which is the bypass arriving through the back door. Three guards now pin it,
each using git's own matcher rather than a hand-rolled glob, plus a literal-source
assertion that the module still reads no filesystem and resolves no SHA of its
own. Sensitivity proven per guard, each red only for its own reason: adding
`.branch-convergence.yml` to `.gitignore` reds the first, `.branch-*.yml` in the
managed block reds the second, a `node:fs` import in the module reds the third.
All three probes reverted; 31 pass.

No `.branch-convergence.yml` is added here — manufacturing repository intent is
exactly what `resolveBase` refuses to do.

### `8cba49fc38ed` (high) — FALSE POSITIVE

The premise is a CI agent running untrusted PR content with repo secrets and
write permissions. No CI job invokes this script:

```
$ grep -rn "sync_pr_branch\|branch_convergence\|sync-pr-branch" .github/
$ echo $?
1
```

The only invocation surfaces in the tree are `taskfiles/dev.yml:173`
(`task push-ready`, a maintainer's local six-step push flow) and the
`/pr:create` command doc at `src/domains/git/pr/create/command.md:163`.
`src/scripts/install-hooks.sh` only *mentions* the module in a comment pointing
the maintainer at `task push-ready`; the hook itself refuses and never merges,
which `tests/scripts/branch_convergence.test.ts` "2.2" already asserts. The
merge and push therefore run on the maintainer's own machine under their own
credentials — there is no `GITHUB_TOKEN`, no workflow write permission, and no
CI agent anywhere on this path.

The wider class the finding gestures at is closed too, and not by assertion:
this tree has **zero** `pull_request_target` workflows — the five grep hits for
that string are all comments recording its absence — so a fork PR reaching any
workflow here is token-capped read-only by GitHub. `self-review-gate.yml` (the
job that produced this very comment) runs on `pull_request` and is
inert-without-secret by design.

### Remaining findings

The other eight on this PR are marked Advisory and are outside this pass, which
covered the blocking-severity security set. Two of them look worth a maintainer
read on their own merits rather than mine: `f8df6af3ba7e` (`remoteSha` trusted
to not lie) and `1ef902de4363` (the `enabled: false` kill switch is surfaced as
BYPASSED by `renderBaseSummary` but writes no audit trail).

PR #1921, opened 2026-09-07
roadmap: a commitment horizon that is decidable, and a reopen that ends

matze4u:
`road-to-limited-commitment-horizon`, 3 of 20. Phases 1 and 2 are prose-complete; the five fixture-gated steps stay open and this PR says exactly why.

## The defect

Choosing a form was implicitly permission to execute a plan for it. Two notes slots already existed and neither was bound to anything:

- `## Predictions` (`src/rules/notes-first-reasoning.md`) carried a prediction with **no subject**, so when an observation came back "the prediction missed" pointed at nothing and reopened nothing.
- `## Decisions` carried `revisit-if`, which nothing reads back.

## What landed

**1.3 — the two fields.** `## Predictions` names the **chosen form** first; `## Decisions` gains `next-commitment`, the boundary the choice authorises work up to. No new notes section and no new file — the round this came from proposed a per-task YAML under `agents/tmp/`, which is a user-only producer surface in this tree.

**1.1/1.2 substance — the horizon is enumerated, and its width scales with reversibility.** The boundary is the nearest of: test result · type or schema inspection · compile · runtime probe · call-site inventory · dry run · browser observation · dependency-contract read. Enumerated on purpose: *"the work the evidence supports"* is a judgement, and the source round showed `next justified commitment` as prose degrading to **"implement solution"** — the starting point wearing a label.

| The action is… | The horizon runs to… |
|---|---|
| a reversible local edit | a whole implementation slice — stopping sooner costs more than it saves |
| stateful or cross-layer | the next boundary from the list |
| irreversible or a public contract | **before** the irreversible step, unless direct evidence already covers it |

**1.4 — the look-ahead ladder is a ceiling, never a quota**, with the quiet reversible case at **one, deliberately**. Always-three-to-five across all variants is 1024 leaves at four branches and depth five. Verified: `grep -rn 'minimum depth|three to five moves'` over `src/rules/`, `src/agent-src/contexts/` and `src/skills/` returns **exactly one hit** — the new ceiling sentence itself. No artifact requires a minimum depth.

**2.1/2.2/2.4 substance — reopen once, on contradiction.** From the candidate list that already exists, never a fresh enumeration, capped at one per candidate; the second contradiction hands over to the retry ladder. A ritual reopen is ceremony and an uncapped one is a loop with better manners. And a rejection whose premise is falsified **is not a rejection** — the candidate is un-evaluated and returns to the set, which is what makes a stated `killed-if` worth carrying at all.

**2.3 — stage 0 on the retry ladder.** Where a candidate list exists, the first failure **reads it before retry 1**. A failure is evidence about the *chosen form*, and retrying the same approach spends the budget without ever reaching the second-best candidate — which is precisely how the mechanism that discovers a form was wrong never routes back to the form. Reading the list is **not an attempt and is not counted**; N=3 is unchanged, and where no list exists the ladder is untouched. This adds a branch; it does not move a budget.

## What is open, and why prose is not enough

**1.1, 1.2, 2.1, 2.2 and 2.4 stay unticked.** Each verifies against a **fixture** — *"a fixture whose chosen form needs a call-site inventory produces a horizon that stops at the inventory"*, *"three fixtures, one per class, produce three different horizon widths"*. Their substance is in the rule as of this change. **Ticking them on the prose alone is the silent-green failure this repository has recorded before**, so they are not ticked.

The roadmap now records what those fixtures actually need, because neither existing layer of `tests/reasoning-layer-eval/` fits: `trigger-fixtures.json` answers *did the right discipline fire*, which is not the question (the question is *where did the run stop*), and the rubric layer scores whole transcripts on 0–3 dimensions rather than asserting a boundary. A horizon fixture needs a third shape — a prompt whose plan has a **known** nearest evidence-producing boundary, plus an assertion that the run stopped at it. That corpus and assertion form is the real remaining work of Phases 1 and 2, not a formality on top of the prose.

**Phase 3** is untouched by design: every step in it reads `road-to-candidate-moves-floor`'s Phase-3 results file as its entry condition, and that measurement has not run.

## Why the three that ARE ticked could be

They verify against **text**, not behaviour, and each was checked rather than assumed: 1.3's field list carries the chosen form and `next-commitment`, and the projection regenerates byte-identically (`condense.sh --changed` → *"Every .md projection matches its source"*); 1.4's grep is above; 2.3's ladder names the candidate-list check as stage 0 with N=3 explicitly unchanged.

## Cost, stated rather than absorbed

`rules_tokens` 65,235 → 66,038 in the meta-pack passport (+803). `notes-first-reasoning` is `type: auto`, so this is **not** always-loaded budget.

Gates green: `lint_roadmap_blockers`, `lint_roadmap_complexity`, `lint_roadmap_ci_steps`, `check_roadmap_trackable`, `check_no_roadmap_refs`, `lint_empty_roadmaps`, `lint_roadmap_later_disposition`, `lint_plan_risk_register`, `check_condensation`, `check_references`.


matze4u:
**Conflict resolved by an owner-delegated drain run — one file, and the resolution is a regeneration rather than a choice.**

This PR was `CONFLICTING`/`DIRTY` with no live session on its branch (`ac-prdrain3` has it checked out; last push 2026-09-08T06:43:27Z). Since it blocks nothing else but was going stale, I merged `origin/main` into it and pushed `bfdb0120f..0bd6fedea`. **No commit was dropped, reordered or force-pushed** — a merge commit on top, nothing rewritten.

**The single conflict was `src/domains/meta/pack.yaml`**, in the generated `token_passport` block:

| Side | `rules_tokens` | `total_tokens` |
|---|---|---|
| this branch | 65643 | 306513 |
| `origin/main` | 65511 | 306381 |
| **regenerated** | **65632** | **306502** |

Neither side was correct for the merged tree — each is a measurement of its own. Taking either would have shipped a passport that describes a tree nobody has. I took `--theirs` and re-ran `generate_pack_manifests`, which produced the third value above; that is the merged tree's actual measurement, and it is between the two as you would expect.

Also re-ran `./agent-config roadmap:progress` after the merge, because the dashboard is generated and a clean auto-merge still leaves it describing neither side's roadmap set. `check_estate_count` green afterwards.

`ci_settle` is running; I will not call this green until it returns a real exit code. The PR now reads `MERGEABLE`.

I did not touch anything else on the branch — its content decisions are its author's.

matze4u:
**CI now runs on this branch, and it reproduces the exact collision that ADR-264 was written about.** Reporting rather than fixing — the fix is a content decision for this PR.

```
measured total   138596 tok (baseline 102520, +36076; ceiling 138490)
❌  per-spawn preamble payload grew past the ratchet: 138596 > 138490 tok.
    net delivered-token change: +178 tok
    grew or arrived (top 1 of 1):
         +178 tok  notes-first-reasoning (rule, grew)
```

**This is not caused by the merge I pushed.** The attribution line names `notes-first-reasoning` — this branch's own rule growth — as the whole of the +178. What the merge changed is only that the conflict is gone, so CI can run at all; the overage was already here and was invisible while the PR was `CONFLICTING`.

**It is the third occurrence of a defect that now has a decision record.** `road-to-a-standing-budget-with-headroom` names three PRs hitting one collision: #1707 (+770) and #1920 (+298) each escaped by migrating prose out of the standing rule, and **this PR cannot** — what remains after migration is two fenced Iron Law blocks, and `preservation-guard` forbids condensing those. Migration was the escape hatch and it does not exist for an Iron Law. (The roadmap recorded 117 residual tokens; the live reading is 178, because `main` moved underneath it.)

**The disposition, from ADR-264** (PR #1940, council quorum 2/2 — they split, and the split was resolved on the dissenting seat's own falsifier):

- The grace ceiling **may not rise**. The two historical raises stand as violations caused by the file's contradictory wording, explicitly not as precedent.
- A 128-token aggregate Iron Law reserve is **designed and deliberately not shipped**, because the deciding seat made enforcement-before-wording an ordering condition: *"Wording alone creates an exploitable trust gap."* The reserve needs protected approval metadata the PR author cannot edit, and no such store exists yet.
- Stated directly in the ADR's Consequences: **this PR does not merge under an exception that does not exist yet.** Until enforcement, approval and fixtures are live, the residual needs a compensating reduction like any other standing growth.

So the honest options here are a compensating reduction elsewhere in the standing payload, or waiting on the reserve's blocker (`iron-law-reserve-needs-protected-approval`). Raising the ceiling is the one move the decision closes off.

The second red, `Node Tests (ubuntu-latest, shard 3/4)`, is still resolving — I will follow up with what it actually says rather than guess.

matze4u:
**Follow-up, as promised: the second red is the same defect, not a second one.**

`Node Tests (ubuntu-latest, shard 3/4)`:

```
× the gate reds on growth past whichever ceiling applies
  > exits 0 under the grace ceiling the CI step passes
FAIL tests/scripts/check_preamble_payload_budget.test.ts
AssertionError: expected 1 to be +0
```

That test asserts the budget gate **exits 0 under the grace ceiling**. This branch is 138,596 against a 138,490 ceiling, so the gate exits 1 and the assertion fails. One cause, two symptoms — the workflow step and the unit test are both reporting the same +178 tok in `notes-first-reasoning`.

Worth stating because it changes the fix rather than adding to it: there is nothing to repair in the test, and no second investigation to run. Whatever resolves the overage clears both checks together.

No third failure — every other check on the branch passes.

